PR #8, opened 2021-06-09
U/lynnej/update radar

rhiannonlynne:
Minor changes to operation of scimaf_dir (mostly in how it writes data into the trackingDb). 
Minor cleanups to comments/captions in metric outputs.
Added usecache=True for SNSL metric slicer (was probably not necessary given nside=16, but in case nside was set higher, this is the correct flag).

### DIFF
--- a/bin/maf/scimaf_dir
+++ b/bin/maf/scimaf_dir
@@ -28,7 +28,7 @@ if __name__ == "__main__":
     run_names = [os.path.basename(name).replace('.db', '') for name in db_files]
 
     for filename, name in zip(db_files, run_names):
-        outDir = name + 'sci'
+        outDir = name + '_sci'
         # Clobber output directory if it exists
         if os.path.isdir(outDir):
             shutil.rmtree(outDir)
--- a/bin/maf/scimaf_dir
+++ b/bin/maf/scimaf_dir
@@ -1,15 +1,15 @@
 #!/usr/bin/env python
-
+import os
+import argparse
 import matplotlib
 matplotlib.use("Agg")
-import os
 import glob
 import shutil
+
 import rubin_sim.maf.batches as batches
 import rubin_sim.maf.db as db
 import rubin_sim.maf.metricBundles as mb
-import argparse
-import os
+import rubin_sim.maf.utils as mafUtils
 
 if __name__ == "__main__":
     """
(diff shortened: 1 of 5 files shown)


bin/maf/scimaf_dir
```diff
@@ -21,21 +21,34 @@ if __name__ == "__main__":
...
+        resultsDb = db.ResultsDb(outDir=outDir)
+        # Set up the metricBundles
         bdict = batches.scienceRadarBatch(runName=name)
-        resultsDb = db.ResultsDb(outDir=name + '_sci')
-        group = mb.MetricBundleGroup(bdict, opsdb, outDir=name + '_sci', resultsDb=resultsDb, saveEarly=False)
+        # Run them, including generating plots
+        group = mb.MetricBundleGroup(bdict, opsdb, outDir=outDir, resultsDb=resultsDb, saveEarly=False)
         group.runAll(clearMemory=True, plotNow=True)
         resultsDb.close()
+        try:
```
yoachim, line 44: I don't think this try block is ever going to execute since we don't have config tables anymore.

yoachim (approved):

bin/maf/scimaf_dir
rhiannonlynne: We don't have the 'info' table? Where is any of the information about the version of the inputs that produced the simulation stored?

rhiannonlynne: The mafUtils.writeConfigs is a utility which calls the fetchConfig method on the OpsimDatabase and then writes the version information into the output MAF directory along with some additional information on when MAF was run and what date it was run on -- creating the Config files that go into those maf output directories.
Maybe copying the info table isn't needed -- it was more useful with opsim3 where you could extract a subset of the 'most useful' information from the Configuration table in the database, but now there really isn't a short set of 'most useful' information lines. If that version information isn't copied, it does imply a greater need to keep the actual survey.database file (which we do anyway), but in theory, the information in the Config files is intended to 
(A) let you identify the precise simulation the MAF outputs were run on, including the version of opsim/input files
(B) track which version of MAF was run to create the output metric data.
The second one of those isn't tracked anywhere else, so is worth preserving.Suppose we could start writing MAF version and date into each of the metric output files, but we're not even keeping those files much of the time, except for the summary statistic information. So - I guess the alternative would be to write the MAF version and date into the MAF metric data files and also into the resultsDb (and if neither of those are created, then the data is considered truly ephemeral and not necessarily repeatable).

rhiannonlynne: Anyway -- it seems like maybe deciding on a better way to keep this information would be useful, but that this is fine for now.

rhiannonlynne:
Deleted branch u/lynnej/updateRadar

bin/maf/scimaf_dir
yoachim: Ah, the docstring on fetchConfig is outdated. Something changed in the migration that broke fetchConfig and it doesn't have a unit test.

PR #19, opened 2021-07-09
gen 2 hourglass plotter with supporting slicer and time use metric

ehneilsen:
I think the hourglass code here is finally in a suitable state for a merge into main. It is not in a final state (e.g. I think there are still same features that should be tweakable in plotDict that aren't yet), but I would rather do these as separate (issue)/branch/merge sets, rather than continuing to do more and more in this branch with nothing in main.

### DIFF
--- a/rubin_sim/maf/metrics/useMetrics.py
+++ b/rubin_sim/maf/metrics/useMetrics.py
@@ -1,0 +1,65 @@
+# imports
+import numpy as np
+
+from rubin_sim.maf.metrics.baseMetric import BaseMetric
+
+# constants
+
+__all__ = ["UseMetric"]
+
+# exception classes
+
+# interface functions
+
+# classes
+
+
+class UseMetric(BaseMetric):  # pylint: disable=too-few-public-methods
+    """Metric to classify visits by type of visits"""
+
+    def __init__(self, noteCol="note", **kwargs):
+        self.noteCol = noteCol
+        super().__init__(col=[noteCol], metricDtype="object", **kwargs)
+
+    def run(self, dataSlice, slicePoint=None):  # pylint: disable=invalid-name
+        """Run the metric.
+
+        Parameters
+        ----------
+        dataSlice : numpy.NDarray
+           Values passed to metric by the slicer, which the metric will use to calculate
+           metric values at each slicePoint.
+        slicePoint : Dict
+           Dictionary of slicePoint metadata passed to each metric.
+           E.g. the ra/dec of the healpix pixel or opsim fieldId.
+
+        Returns
+        -------
+        str
+            use at each slicePoint.
+        """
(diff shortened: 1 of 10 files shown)


rubin_sim/maf/metrics/useMetrics.py
```diff
@@ -0,0 +1,65 @@
+# imports
+import numpy as np
+
+from rubin_sim.maf.metrics.baseMetric import BaseMetric
```
rhiannonlynne, line 4: Can you make this a relative import? 
from .baseMetric import BaseMetric
(I believe relative imports are preferred, but please comment if that's not correct)

ehneilsen: I've always understood the opposite. From PEP8: "Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path)... However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose:... Standard library code should avoid complex package layouts and always use absolute imports. Implicit relative imports should never be used and have been removed in Python 3."

The google python style guide is even more extreme, and does not allow relative imports at all (see section 2.3 of the [google style guide](https://google.github.io/styleguide/pyguide.html)).

rhiannonlynne: I see. I can see how you might say our import line is pretty long/verbose to make relative preferred here, but also it's not that long, and given that the absolute import is preferred - glad I asked.

timj: In DM we are more than happy to use relative imports within a single package...

rhiannonlynne (approved):
Sorry for the delay in writing this down - can you rebase on main, then use rubin_sim.utils.ddf_locations to get the locations of the DDF fields instead of the DEEP_COORDS hard-coded below? (and potentially let the user override those if desired, say if they're looking at a different opsim from a different generation).

rubin_sim/maf/plots/hgPlotters.py
```diff
@@ -0,0 +1,1200 @@
...
+from rubin_sim.utils.riseset import riseset_times
+
+# constants
+
+# The DDF fields are identified here (instead of pulled from config
+# file somewhere)
+# Thus need to be updated if field locations change --
+# these are used to set transit lines
+
+DEEP_COORDS = OrderedDict(
```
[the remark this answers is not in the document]
ehneilsen: Okay, just checked in the change.

rhiannonlynne (approved):

rhiannonlynne:
Rebased and merged u/neilsen/hourglass, deleted branch.

PR #25, opened 2021-07-28
U/lynne/cleanup sci

rhiannonlynne:
This just tidies up some of the metrics and their summary stats. 
Some of it is cleaning up the docstrings and removing unneeded warts. 
Some of it is refactoring some code. 
Some of it was just changing the summary stats for light curve related metrics so that the output is a little more extensive.
Tried to understand the SNSL metric better - it seems to be doing what it says it should be doing, so now I think it's over to DESC to let us know what's going on, if the metric results are what they expected.


rhiannonlynne:
One thing that's debatable -- the collapse_night utility. 
I modified it so that it was more readable for me .. and it's very fractionally faster. BUT it does lose the "carry along all other data in the data slice" component that the old version had .. but then again, the other information in the Dataslice wasn't being used. So to me it's a wash-up and I'd go with this one to avoid the possibility of messing up DataSlice without realizing it in the future. But I could be convinced otherwise.

PR #33, opened 2021-08-04
update to newer throughputs

yoachim:
Updated throughputs to throughputs v1.5


rhiannonlynne:
Shouldn't this to be to update the throughputs to v1.7? 
https://github.com/lsst/throughputs/blob/master/baseline/README_SOURCE.md

yoachim:
Ah, yes. And it was. Looks like updating commit messages involves force pushes. If it matters, the easiest is probably to make a trivial change to the readme or something and say there that we're up to 1.7.

PR #34, opened 2021-08-05
fixing generate_ss script because Lynne broke it yet again

yoachim:


rhiannonlynne:
Yeah, I haven't found it useful to just tell the script to run on all database files in a given directory yet - sorry for the glob miss. 
(let me also note that running in split file mode, which I doubt you will do anyway, won't work well for more than one database at a time yet .. it's possible to do, but wasn't how I set it up since I was really trying to get the processing time for one database down to something under two hours instead of over a day).

PR #43, opened 2021-08-18
U/lynnej/update one dslicer

rhiannonlynne:
Fix oneDSlicer setup (or re-setup) problems, and simplify the bin specification. 
Also updates the unit test, removing old references to where 'bins' could be set by a single integer (being the number of bins  .. now it's just an array).

### DIFF
--- a/rubin_sim/maf/slicers/oneDSlicer.py
+++ b/rubin_sim/maf/slicers/oneDSlicer.py
@@ -43,6 +43,9 @@ class OneDSlicer(BaseSlicer):
         self.bins = bins
         # Forget binmin/max/stepsize if bins was set
         if self.bins is not None:
+            if binMin is not None or binMax is not None or binsize is not None:
+                warnings.warning(f'Both bins and one of the binMin/binMax/binsize was specified. '
+                                 f'Using bins ({self.bins} values only.')
             self.binMin = self.bins.min()
             self.binMax = self.bins.max()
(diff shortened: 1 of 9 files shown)


rubin_sim/maf/slicers/oneDSlicer.py
```diff
@@ -12,81 +12,94 @@
...
         self.columnsNeeded = [sliceColName]
         self.bins = bins
-        self.binMin = binMin
-        self.binMax = binMax
-        self.binsize = binsize
-        if sliceColUnits is None:
-            co = ColInfo()
-            self.sliceColUnits = co.getUnits(self.sliceColName)
+        # Forget binmin/max/stepsize if bins was set
+        if self.bins is not None:
```
ehneilsen, line 45: It might be worth raising an error (or at least a warning) if a user provides both explicit bins, and also binMin etc. such that the later gets ignored.

ehneilsen (approved):
Looks good, just one comment (see below).

rubin_sim/maf/slicers/oneDSlicer.py
rhiannonlynne: Good point, always good to let the user know if something potentially unexpected happened.

ehneilsen (commented):
Okay, should have approved this ages ago...

rhiannonlynne:
Thanks!

PR #51, opened 2021-09-15
U/yoachim/short survey

yoachim:
Expand the short_survey detailer so it can be more

### DIFF
--- a/rubin_sim/maf/batches/scienceRadarBatch.py
+++ b/rubin_sim/maf/batches/scienceRadarBatch.py
@@ -214,7 +214,7 @@ def scienceRadarBatch(colmap=None, runName='opsim', extraSql=None, extraMetadata
     bundleList.append(bundle)
 
     displayDict['caption'] = 'Microlensing events with crossing times between 30 to 100 days.'
-    slicer = generateMicrolensingSlicer(min_crossing_time=10, max_crossing_time=30)
+    slicer = generateMicrolensingSlicer(min_crossing_time=30, max_crossing_time=100)
     metric = MicrolensingMetric(metricName='30 to 100 day Microlensing')
     bundle = mb.MetricBundle(metric, slicer, extraSql, metadata=extraMetadata,
                              runName=runName,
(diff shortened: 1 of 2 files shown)


rubin_sim/maf/batches/scienceRadarBatch.py
```diff
@@ -214,7 +214,7 @@ def scienceRadarBatch(colmap=None, runName='opsim', extraSql=None, extraMetadata
     bundleList.append(bundle)
 
     displayDict['caption'] = 'Microlensing events with crossing times between 30 to 100 days.'
-    slicer = generateMicrolensingSlicer(min_crossing_time=10, max_crossing_time=30)
+    slicer = generateMicrolensingSlicer(min_crossing_time=30, max_crossing_time=100)
```
rhiannonlynne, line 217: Good catch.

rubin_sim/scheduler/detailers/short_survey.py
```diff
@@ -18,10 +18,15 @@ class Short_expt_detailer(Base_detailer):
         The short exposure time to use.
     nobs : float (2)
         The number of observations to try and take per year
+    night_max : float (None)
+        Do not apply any changes to the observation list if the current night is greater than night_max.
+        XXX--note, in theory, detailers should probably be expanded to include their own basis functions.
+    in_a_row : int (1)
+        How many short observations to do in a row.
```
rhiannonlynne, line 25: Maybe call this 'sequential' or 'sequence'? 
I guess the cadence notes and white papers on short exposures don't specify whether the short exposures have to be at the same or different times. Might try shooting Will a note to check into if they have any preference. Agreed it's probably more efficient to do sequential if there are more than one short exposure to be done overall.

rubin_sim/scheduler/detailers/short_survey.py
```diff
@@ -18,10 +18,15 @@ class Short_expt_detailer(Base_detailer):
         The short exposure time to use.
     nobs : float (2)
         The number of observations to try and take per year
+    night_max : float (None)
+        Do not apply any changes to the observation list if the current night is greater than night_max.
+        XXX--note, in theory, detailers should probably be expanded to include their own basis functions.
```
rhiannonlynne, line 23: Can you move this comment out of the documentation that shows up in the compiled documentation and maybe into a # comment in the code (it seems more like a reminder for development than a comment to communicate to users)?
This might be worth bringing up in a more general idea -- how are detailers separate from surveys, what do they have in common, if they have basis functions should they also have other properties of surveys?

rhiannonlynne (approved):

PR #55, opened 2021-09-20
U/lynnej/docs

rhiannonlynne:
Add GitHub workflows for unit tests and for documentation.


rhiannonlynne:
This PR adds a basic workflow for the repository to run unit tests and then build and push the documentation to rubin_sim.lsst.io. 

Some changes were made in order to support running the unit tests without having to have a large amount of data. 
I ended up with the 'minimal' set of site_models, throughputs, sky brightness, skybrightness_pre and the tests data directories. 

In order to drop "maps", I dropped tests of the file size of the dust maps and stellar maps classes (these were hold-overs from when we had git-lfs, so not really needed anymore anyway). Then I also added configuration paths to the DustMaps and StellarMaps classes (to allow putting small versions of these into 'tests' data directory for use in testing these Maps classes). 

In order to drop MAF, I had to drop unit tests for the SN metrics and for building the ScienceRadarBatch metric -- both of these read in large data files regarding light curves and throughputs gamma matrices. These tests *do* still run if the data is available; they just test to see if it's there first. The other change I had to do in order to drop 'maf' data was to add a copy of the camera FOV map to tests -- and then make the camera FOV map path configurable. The easiest (and easier for longer term use) for that was to add a separate LsstCameraFootprint class in the utils directory that operates in much the same way as the individual pieces of code we previously had in movingObjects and in maf worked (the class can be configured to expect either degrees or radians, as these pieces of code use different options). There is an example of usage in https://github.com/lsst/rubin_sim_notebooks/blob/u/lynnej/add_lsstCameraFootprintDemo/utils/Camera%20Footprint%20Demo.ipynb

There was some other configuration changes on where to look for data, but those were the primary changes. `test_CoreScheduler` now uses the small seeing.db from tests although it is not really necessary since we ended up keeping site_models .. could go either way, but it's fine as-is. 
The Fields_database class was modified a bit to only read in the data from the database when it's needed rather than on init .. but this could be reversed too, especially as we ended up keeping the site_models data and I fixed the usage in MAF which required that database to be present on import of the stackers (even if not actually used). (mostly I just didn't bother swapping it back .. could do though).

rhiannonlynne:
As far as the GitHub Actions workflow: 
both ubuntu and macOS versions of the unit tests (and then a build and push of the documentation) are run on push to main, PR, or on release creation. The workflow can also be manually triggered. 

there is also a workflow which downloads all of the rubin_sim_data and runs all unit tests, but is only manually triggered.

rhiannonlynne:
Oh - and I added a requirements.txt file so the install steps could be:
```
git clone git@github.com:lsst/rubin_sim
cd rubin_sim
conda create -n rubin   ### optional (but recommended)
conda activate rubin    ### optional (if new environment created above)
conda install -c conda-forge --file=requirements.txt
pip install -e .
```
which .. is longer than previous, but we wouldn't need to edit the requirements list each time (just edit in the requirements.txt file).

The download data script can also use the flag --force to force an updated download of the data files.

rubin_sim/utils/cameraFootprint.py
```diff
@@ -21,50 +20,54 @@ class LsstCameraFootprint(object):
...
-            Ephemerides for the objects, with RA and Dec as 'ra' and 'dec' columns (in degrees).
-        obsData : np.recarray
-            Observation pointings, with RA and Dec as 'ra' and 'dec' columns (in degrees).
-            The telescope rotation angle should be in 'rotSkyPos' (in degrees), and the time of each
-            pointing should be in the 'expMJD' column.
-        rmax : float (2.1)
-            The maximum radius from the center of the field to consider
-        epoch: float, optional
-            The epoch of the ephemerides and pointing data. Default 2000.0.
+        obj_ra : `np.ndarray`
```
ehneilsen, line 41: I don't know if the fancy linking done in creating of the documentation will be confused by using the `np.recarray` abbreviation rather than the full `numpy.recarray`. This should be checked.

tests/skybrightness_pre/test_ZernikeSky.py
```diff
@@ -47,6 +47,8 @@ def test_bulk_zernike_fit(self):
         self.assertGreater(reread_zernike.shape[1], 20)
         self.assertEqual(tuple(reread_zernike.index.names), ("band", "mjd"))
 
+        test_out_dir.cleanup()
```
ehneilsen, line 50: This shouldn't be necessary. TemporaryDirectory objects automatically clean themselves up when the go out of scope. An explicit cleanup should only be necessary if you want to delete the directory before the instance of TemporaryDirectory goes out of scope (or is explicitly deleted).

rubin_sim/utils/cameraFootprint.py
rhiannonlynne: I think that I figured out maybe this problem I saw is due to not importing things in the conf.put of the doc build stage. I’ll check.

tests/skybrightness_pre/test_ZernikeSky.py
rhiannonlynne: You’re right that it cleans up itself, but it also gives a warning while doing so. This removed the warning message.

yoachim (approved):

ehneilsen (approved):
Wow, this looks like it was a lot of fairly tedious work.  I think having the generated documentation online will be a big help, though.

Aside from the pretty trivial comments I made, it looks good to me.

rhiannonlynne:
The unit tests pass, the documentation looks reasonable in my built version, and the push has succeeded in previous attempts. I think this last failure was just a connection issue on the lsst.io end .. hopefully. 
Will merge and check on how it builds from main.

rhiannonlynne:
Added output schema from u/yoachim/document_schema

rubin_sim/utils/cameraFootprint.py
rhiannonlynne: It still doesn't work right, but neither does numpy.ndarray .. numpy.recarray seems like it's also failing, but I'm less sure. 
Going with this for now, but should check in with dm-docs.

PR #56, opened 2021-09-25
Make documentation easier to read and update

rhiannonlynne:
Moves around python API pages (giving more room to add text on each module page). 
Fixes the links in the metricList page (updating the metricList.py code).
Fixes numerous docstrings (primarily for indentation) and adds even more missing __all__ methods. (note that if __all__ is not included, automodapi will try to build an api documentation stub for imported methods .. not really what we want, so please do add __all__ lines at the top of each file for what should be available for import from the file).

### DIFF
--- a/.github/workflows/python-docs-only.yml
+++ b/.github/workflows/python-docs-only.yml
@@ -1,0 +1,54 @@
+name: Build and Push Documentation
+
+on:
+  # Trigger the workflow manually (otherwise runs automatically after unit tests)
+  workflow_dispatch:
+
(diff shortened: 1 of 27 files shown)


rhiannonlynne:
After having the doc workflow broken by a failed upload a few times, I added a docs-only workflow as well. This is only triggered manually (go to 'actions', click on the "Build and Push Documentation" workflow available after this is merged to main, then choose to Run workflow .. once it's available, you will be able to run the workflow on any branch).

PR #58, opened 2021-09-28
adding agn structure function

yoachim:
Nice little structure function metric from Weixiang Yu.

### DIFF
--- a/rubin_sim/maf/metrics/__init__.py
+++ b/rubin_sim/maf/metrics/__init__.py
@@ -37,3 +37,4 @@ from .seasonMetrics import *
 from .scalingMetrics import *
 from .useMetrics import *
 from .brownDwarfMetric import *
+from .agnstructure import *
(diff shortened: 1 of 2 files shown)


rhiannonlynne (approved):

rhiannonlynne:
Looks good. unit test? (it's ok without it, but something we should consider asking for more often)

rhiannonlynne:
Oh one more: should this be added to the ScienceRadar? 
(we may want to make more sub-science area batches, instead of just adding directly into science radar each time .. it would make subsets metrics in the set easier to find and run)

yoachim:
> Oh one more: should this be added to the ScienceRadar? (we may want to make more sub-science area batches, instead of just adding directly into science radar each time .. it would make subsets metrics in the set easier to find and run)

I think we should add it to ScienceRadar when we get the AGN folks to converge on what they agree a good summary metric would be. I imagine it would be something like "total area where the structure function error is less than 0.05 mags for a g=22 AGN", but we need them to say what limits and filters make the most sense.

PR #60, opened 2021-10-01
Make sure not to crash if directory is current directory

rhiannonlynne:
labelVisits creates a new copy of the opsim database, in order to change the proposal labels. 
The new database is created with a new name .. the way the name was created would crash if this was being run from the current working directory, instead of one directory up.
This PR fixes that bug.


ehneilsen (commented):
Looks reasonable.

PR #74, opened 2021-10-19
U/yoachim/col clear

yoachim:
Expose the column registry so unit tests can clear it and not have registered columns persist.


rhiannonlynne:
One comment - rebase/merge the "Hourglass patch" PR first maybe? (because this one was branched from there, not main)

PR #81, opened 2021-11-12
Add NQSO metric 

rhiannonlynne:
Contributed from @rjasseff this metric adds an 'expected number of QSO' metric. 
The cadence notes regarding QSO cover three metrics: 
* coadded depth after extragalactic extinction compared to quasar colors and finds little difference between simulations (i.e. this is essentially looking at filter balance)
* coadded depth in one band to evaluate total number of quasars (i.e. this is essentially looking at number of visits vs area trades)
* short-term intervals of observations between bands to evaluate single-image color and variability determination possibilities (i.e. cadence, basically)

For MAF, it seemed reasonable to shift these to:
* total number of quasars (from i band) -- IMPLEMENTED HERE,
*  total number of quasars with well-estimated colors or photometric redshift from coadded images (combine point one and two), -- TBD and then 
*  potentially combine the total number of quasars with the number of those which would have gotten short-term color/variability information (if it's possible to determine what timescales are needed in order to measure this last piece of information)? TBD, but maybe covered by SFError metric and/or AGN Time Lags metric


rhiannonlynne:
In order to pull this in from the other repo, test and refactor it a small amount (primarily to point it to the rubin_sim_data files), I'm making a PR to a branch, and then we'll merge the branch to main after these small changes.

rhiannonlynne:
Ok .. actually the fork included a lot of extra data files and notebooks which I do not want to put in the rubin_sim git history (as doing this kind of thing will eventually bloat the repository, and we don't need 'how to generate the data files' here). I'll make an lsst_sims repository to hold some of this maybe? For now, I just want the metric itself and the data files which I'll put elsewhere.

rhiannonlynne:
Building this on the u/lynnej/add_qsoCounts

PR #84, opened 2021-11-19
YoungStellarObjects Metrics 

lprisinzano:
We require to incorporate the YoungStellarObjects Metrics

rubin_sim/maf/mafContrib/211116_yso_3D_90b.py   into MAF. Please read the README.md file

Justifications are given in the attached pdf file
[2101116_YSO_inputs.pdf](https://github.com/lsst/rubin_sim/files/7571344/2101116_YSO_inputs.pdf)


rhiannonlynne:
Hi @lprisinzano thank you for the PR!
Can I ask for a few refactoring favors before approval? 

* We have a VolumeSumMetric and DustValues class available in rubin_sim - can you please just import the existing classes rather than adding new versions of them? 
* can you split the actual metric class itself out into its own file please? (or a file with other basic MAF. building blocks is fine)
* generally, we are not encouraging scripts to run the metric itself being added here .. if you have a notebook illustrating the usage, could you please send it in a PR to rubin_sim_notebooks? (https://github.com/lsst/rubin_sim_notebooks) If instead you want to show us the usage in a script as you have here, that is fine -- we will likely end up deleting the file, setting up the configuration in the 'batches' and then adding a notebook so we can see how things are working (just so there are no surprises).

lprisinzano:
Hi @rhiannonlynne,
many thanks to have accepted my PR. 

Regarding the DustValues class: the first version of the YSO metrics used the LSST dust extinction map (DustValues class), but this was based on the Schlegel et al. 1998 dust map that is bidimensional (2D). This means that each E(B-V) is based only on the coordinates (l, b) and  that  the returned E(B-V) values are integrated at the maximum distance. But E(B-V) values  actually increase with distance and using a 2D map we would have bias results . After a great effort by Will Clarkson and Alessandro Mazzi, we now have a more realistic  extinction 3D map based on  (l,b,dist). Unless the LSST class DustValues has not been upgraded to a 3D map,   we strongly suggest to adopt the 3D extinction map. 

I have a metrics written as a Notebook and I will include it in the rubin_sim_notebooks.

lprisinzano:
However, Will Clarkson and Alessandro Mazzi are working to making the extinction map MAF-compatible

rhiannonlynne:
Ah, it was not obvious that it was different than the DustValues class we already have. Could the name of the class be changed so that this is not a collision?

lprisinzano:
Hi Lynne, no problem, we can change the class name. We are working on this. Thank you again.

lprisinzano:
pdate version of the Young Stellar Objects metrics. No additional software dependencies are required since the code automatically downloads the repository and the 3D extinction map.  The Dust Map name has been changed to avoid collisions with the default LSST MAF Dust map.  The metrics can be tested with a Notebook using the following commands:

import YoungStellarObjectsMetric
YoungStellarObjectsMetric.run_examples_datalab()

The notebook demonstrating the use of the metrics is  added to rubin_sim_notebooks.
Please, contact  loredana.prisinzano@inaf.it for  updates  about the process of integration of this metrics into MAF and
if you have any questions or need further info.
~

PR #91, opened 2021-12-02
U/yoachim/goodseeing2

yoachim:
Add a basis function so we can include a desire to collect images in good seeing conditions.

### DIFF
--- a/rubin_sim/scheduler/basis_functions/basis_functions.py
+++ b/rubin_sim/scheduler/basis_functions/basis_functions.py
@@ -3,6 +3,7 @@ from rubin_sim.scheduler import features
 from rubin_sim.scheduler import utils
 from rubin_sim.scheduler.utils import int_rounded
 import healpy as hp
+import matplotlib.pylab as plt
 from rubin_sim.skybrightness_pre import M5percentiles
 import warnings
 from rubin_sim.utils import _hpid2RaDec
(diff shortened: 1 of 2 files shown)


ehneilsen:
Are we trying to move to following PEP8 naming conventions? It looks like the new changes are consistent with current usage in rubin_sim, but quite far from PEP8, which specifies that classes are all CapsCase (words capitalized, no underscores), while variables, methods and function names are all snake_case (no capitals, words separated by underscores). Whether this is a problem or not just depends on what we want to do. Personally, I find our failure to follow the normal conventions makes the code harder to read.

ehneilsen:
Do we intend to specify a number of visits with delivered seeing better than some value, or a number of visits taken when the zenith seeing is better than some value? I'm not sure, but it looks like the code does the former, but the latter is probably what we want. I don't think we want to observe a field at high airmass because we need a good seeing image of that field, and the zenith seeing is better than the cutoff.  Cutting on the atmospheric seeing will work most of the time, because either the high airmass won't push it over the limit or the field is at low airmass anyway, but I expect that sometimes it will be important.

yoachim:
Yeah, the plan is we'll go through and fix the naming conventions. I think going though and doing it in a massive change will be better than piecemeal. 

The `conditions.FWHMeff` is a HEALpix map of the seeing at each healpixel including filter and airmass info, so it's not going to accidentally go to high airmass just because the zenith seeing is good.

ehneilsen (approved):

tribeiro:
Please, rebase before merging to main!

PR #93, opened 2021-12-02
Restore setup_requires in setup.py

tribeiro:


yoachim (approved):

rhiannonlynne:
Hi Tiago - looks like you're removing the setup requires from the setup.py file? 
That line was added by @beckermr to enable setuptools_scm to be included in the git archive tarballs. 
Is it causing problems with the builds in the lsst-ts Jenkins?

tribeiro:
I am actually adding a line that was removed

tribeiro:
Yes, it's causing the conda build to fail

rhiannonlynne:
I'm super confused. Matt's PR looked like it *added* that line?

tribeiro:
let me see if I can find a way to co-exist this with the conda build...

tribeiro:
> I'm super confused. Matt's PR looked like it _added_ that line?

Yes, you are right sorry...

beckermr:
add `setuptools_scm_git_archive` into your env

I should have done that. Sorry!

rhiannonlynne:
@tribeiro - @beckermr is setting up a Conda-forge build of rubin_sim https://github.com/conda-forge/staged-recipes/pull/17146/files

tribeiro:
It is part of the conda package dependency but not as a "setup dependency".. I think I can fix it by adding it as a setup dependency.. hold on

beckermr:
see here: https://github.com/lsst/rubin_sim/pull/94

tribeiro:
I'm closing this PR.. I'll patch the conda recipe to add this new dependency...

rhiannonlynne:
Sorry about the unexpected change, Tiago. I didn't realize it would crash the lsst-ts conda builds, and while I can see if they fail, I don't have much transparency into why they're failing (at least in this case). 
Longer term, DM hosts all of its packages on conda-forge and I suspect that's where system performance packages would end up if they're conda-installable .. keeping two channels/packages open is fine, and might be a super reasonable course of action so that you can run the unit tests in the lsst-ts environment (which I think you're doing?).

@beckermr here is where I think the lsst-ts conda recipe can be found: 
https://github.com/lsst-ts/ts_recipes/blob/develop/rubin-sim/conda/meta.yaml

tribeiro:
No problem at all!

I think we can retire the tssw build once we have it on conda-forge...

beckermr:
Yeah that recipe is wrong anyways. It vendors setup time deps

tribeiro:
@beckermr , what do you mean?

beckermr:
Ack I was wrong. Sorry!

beckermr:
This package is on conda-forge now, so feel free to pull from there.

tribeiro:
Nice! thank you!

PR #95, opened 2021-12-03
Update healpixSubsetSlicer

rhiannonlynne:
Add sid to all slicers, make metricBundle identify points in metric values using 'Sid' instead of a counter.
Then update the iteration in healpixSubsetSlicer, so it *only* iterates over the healpix points in the subset (instead of iterating over all points and only returning data for the points in the subset).
This makes the healpix subset slicer much more usable for something like DDF fields, while maintaining a full healpix map of metricValues (most of them masked) -- which lets you use things like hp.gnomview without having to rebin the pixel values (which was introducing additional granulation).

### DIFF
--- a/rubin_sim/maf/batches/timeBatch.py
+++ b/rubin_sim/maf/batches/timeBatch.py
@@ -331,7 +331,8 @@ def timeGaps(colmap=None, runName='opsim', nside=64,
     return mb.makeBundlesDictFromList(bundleList)
 
 
-def seasons(colmap=None, runName='opsim', nside=64, extraSql=None, extraMetadata=None):
+def seasons(colmap=None, runName='opsim', nside=64, extraSql=None, extraMetadata=None,
+            slicer=None):
     """Generate a set of statistics about the length and number of seasons.
 
     Parameters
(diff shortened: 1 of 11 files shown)


rhiannonlynne:
The goal of this ticket is eventually to include DDF metrics -- but I haven't finished that part yet (see https://github.com/rhiannonlynne/notebooks/blob/master/DDF%20Nvisits.ipynb for an example of some of what I was thinking, at least for the visits over the DDF region .. the example there doesn't do the plotting how I'd like to be able to, but that's because I didn't rerun it yet and there may be some tweaking to do to the healpixSkyMap plotting function still. It *should* work). 
Anyway, I don't have the "rest of the story" in place yet, but this is the beginning and I'd like to merge this before we blacken the rest of the codebase. 
Could you please have a look and send back comments?

yoachim:
Looks fine to me. We could take this opportunity to just deprecate and remove `OpsimFieldSlicer`.

rhiannonlynne:
Yeah, I thought about removing the OpsimFieldSlicer .. I think I need to keep it for right now, until I can make sure I can generate appropriately interesting movies using the healpix slicer, but then might be good.

PR #97, opened 2021-12-03
Young Stellar Objects metric

rhiannonlynne:
From Loredana: 

We require to incorporate the YoungStellarObjects Metrics

rubin_sim/maf/mafContrib/211116_yso_3D_90b.py into MAF. Please read the README.md file

Justifications are given in the attached pdf file
2101116_YSO_inputs.pdf

Regarding the DustValues3d class: the first version of the YSO metrics used the LSST dust extinction map (DustValues class), but this was based on the Schlegel et al. 1998 dust map that is bidimensional (2D). This means that each E(B-V) is based only on the coordinates (l, b) and that the returned E(B-V) values are integrated at the maximum distance. But E(B-V) values actually increase with distance and using a 2D map we would have bias results . After a great effort by Will Clarkson and Alessandro Mazzi, we now have a more realistic extinction 3D map based on (l,b,dist). Unless the LSST class DustValues has not been upgraded to a 3D map, we strongly suggest to adopt the 3D extinction map.

I have a metrics written as a Notebook and I will include it in the rubin_sim_notebooks.

Update version of the Young Stellar Objects metrics. No additional software dependencies are required since the code automatically downloads the repository and the 3D extinction map. The Dust Map name has been changed to avoid collisions with the default LSST MAF Dust map. The metrics can be tested with a Notebook using the following commands:

import YoungStellarObjectsMetric
YoungStellarObjectsMetric.run_examples_datalab()

The notebook demonstrating the use of the metrics is added to rubin_sim_notebooks.
Please, contact loredana.prisinzano@inaf.it for updates about the process of integration of this metrics into MAF and
if you have any questions or need further info.


rhiannonlynne:
I'm creating this new PR to a branch on rubin_sim, so that we can clean it up here before merging onto main (removing some scripts and making sure the documentation builds, etc.), without having to make the contributors do this work.

PR #100, opened 2021-12-10
U/yoachim/update 21 defaults

yoachim:
* Add virgo cluster to the default footprint. 
* remove SDO from default solar system script
* fix deprecation warning


rhiannonlynne:
Before we run many runs with this as the baseline footprint, we should run it past the SCOC. 
(a comparison of the metric outputs from the baseline v2.0 and the baseline v2.0 + virgo cluster minisurvey, with a small amount of text, should do it). 
If we're not running a lot of runs that will be part of the v2.1 set, I think it's fine to add this though.

rhiannonlynne (approved):

PR #104, opened 2021-12-15
remove column testing

yoachim:
Blacken the code and remove the column testing that uses `numLimit=1`.


rhiannonlynne:
Ugh - can you add 'requests' to the requirements.txt file? That should let the unit tests pass. 
I think it must have been a requirement subsumed by something else previously, and now it is exposed as missing.

rhiannonlynne:
Otherwise, looks fine.

rhiannonlynne (approved):

rhiannonlynne:
Edit suggestion - I just put in a PR to update the requirements.txt file, so you could approve and merge that one, then rebase on that instead.

PR #106, opened 2021-12-16
speed up scripted surveys

yoachim:
Simplify the scripted surveys a bit to get some speedup when there are a lot of them.

### DIFF
--- a/rubin_sim/scheduler/surveys/scripted_surveys.py
+++ b/rubin_sim/scheduler/surveys/scripted_surveys.py
@@ -65,13 +65,18 @@ class Scripted_survey(BaseSurvey):
                     detailer.add_observation(observation, **kwargs)
                 self.reward_checked = False
 
-                # find the index 
-                indx = np.searchsorted(self.obs_wanted['scripted_id'], observation['scripted_id'])
+                # find the index
+                indx = np.searchsorted(
+                    self.obs_wanted["scripted_id"], observation["scripted_id"]
+                )
                 # If it matches scripted_id and note, mark it as observed and update scheduled observation list.
-                if self.obs_wanted['scripted_id'][indx] == observation['scripted_id']:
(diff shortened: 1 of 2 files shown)


rubin_sim/scheduler/surveys/scripted_surveys.py
```diff
@@ -53,30 +65,13 @@ def add_observation(self, observation, indx=None, **kwargs):
...
-                        # Log it as observed.
-                        self.obs_wanted["observed"][match] = True
-                        self.scheduled_obs = self.obs_wanted["mjd"][
-                            ~self.obs_wanted["observed"]
-                        ]
-                        break
+                # find the index 
+                indx = np.searchsorted(self.obs_wanted['scripted_id'], observation['scripted_id'])
+                # If it matches scripted_id and note, mark it as observed and update scheduled observation list.
+                if self.obs_wanted['scripted_id'][indx] == observation['scripted_id']:
```
tribeiro, line 71: I think you can merge these 2 `if` statements into one no?

```
if self.obs_wanted['scripted_id'][indx] == observation['scripted_id'] and self.obs_wanted['note'][indx] == observation['note']:
...
```

tribeiro (approved):
Code looks good, I have only one small comment inline and one general comment.

The observation object has an `ID` and others `<something>_id` items already (`field_id`, `survey_id` and `block_id`). Would it be possible to use one of those instead of adding a new column to observation?  Maybe you could use `field_id` for this purpose?

I'm thinking in terms of memory footprint (this now adds a new integer field to all observations) but also in terms generality. 

Anyway, just something to think about.

rubin_sim/scheduler/surveys/scripted_surveys.py
yoachim: Yeah, I'd like to deprecate `field_id` and `survey_id` (and `block_id` can be handy, but isn't critical).

PR #107, opened 2021-12-17
U/yoachim/db prune

yoachim:
• Removed Database class and dbObj. Replacing with thin wrapper around pandas.read_sql. Put in a stub so the code is compatible with old notebooks
• Removed Completeness metric
• Uncovered issues with test_hgPlotters, I don't think it was actually running before (was getting no metric values). Set them to skip for now
• Removed OpsimFieldSlicer. Replaced it with UserPointsSlicer in unit tests where applicable


rhiannonlynne:
So if I'm interpreting these database-related changes correctly, the general idea is that 
* "opsimDatabase" is no longer a real object
* all the work of getting observations from the database happens in getSimData only
* we're dropping easy support for group by, but we could do this by directly querying the opsim database and turning the result into a numpy recarray. (now I am pretty happy we didn't do automatic column name substitution, but left the names as in the opsim database outputs).

There is some muddiness in what we're passing to the pandas.read_sql method. 
When it is a string, it could be either a sqlalchemy connection URL or it could be just the filename (but these will not be interchangeable, and the current test would do the wrong thing if a sqlalchemy connection URL is passed). 

Connecting to the database each time rather than opening and saving the connection is slower and we should probably open and save that connection in the OpsimDatabase. It looks like the current implementation might end up pushing users to using the string and then opening/closing each time they call getSimData. 

You could make the OpsimDatabase open a sqlalchemy engine, and save the engine. Then pass the engine to pd.read_sql in order to get the database values. With the sqlalchemy engine, you can also find the table names, to avoid trying one and then the other in the try: except block here (especially where you're just catching all exceptions, and would be reporting the incorrect failure (based on the SummaryAllProps attempt) if something went wrong). 
This would also let you push the connection string determination stuff back into a part of code which is more self-contained in the database area, rather than being in utils (and you'd only have to do it once, instead of each time data is gathered from the opsim output).  Keeping the ability to create a more complex URL doesn't have to mean that we have to build it ourselves in a way that was as complex as in the older dbObj.py. We could assume that most people will just provide the dbfilename, and then we turn it into a sqlite URL, and that if someone wants to do anything else, they will provide the complete URL. (but probably best to do this by setting two kwargs and letting people use one or the other, rather than trying to test if it's a filename string or a fully qualified URL or an actual URL object).

rhiannonlynne:
Another benefit to creating and using a sqlalchemy engine as 'what is the opsimDatabase' class, is that you would then be able to determine and save the table names:

```
from sqlalchemy import create_engine
from sqlalchemy.engine import make_url
from sqlalchemy import inspect

url = make_url("sqlite:///" + databasefile)
eng = make_engine(url)
inspector =

rhiannonlynne:
(sorry - close/reopen was button pushing mistake)

yoachim:
Ah, I clearly forgot to finish that docstring. There's no database URL support. It's just a filename to a sqlite3 file, or a full blown connection object. 

I added in a kwarg to make it possible to specify the sql query, so now if folks need "group by" or other fancy things they can use that. 

I was curious, saving and reusing a connection object was slower than just making a new connection each time:  https://github.com/yoachim/21_Scratch/blob/main/conn_time_test/conn_time_test.ipynb
If anyone wants to generate the connection object they can, but there's no penalty just passing the filename.

I like the inspector stuff. Added that in to avoid the bad try/except block.

rhiannonlynne:
> I was curious, saving and reusing a connection object was slower than just making a new connection each time: https://github.com/yoachim/21_Scratch/blob/main/conn_time_test/conn_time_test.ipynb If anyone wants to generate the connection object they can, but there's no penalty just passing the filename.


Can you do me a favor and try that again, in a fresh notebook perhaps and perhaps reversing the order of the operations of the cells? 
I tried it myself too and found quite different results. (had to use baseline_v2.0_10yrs though)

Screen Shot 2021-12-17 at 6 11 35 PM

rhiannonlynne:
The last two columns in the showman pages (maf version and date) are generated from the writeConfig script running & outputting a maf summary config file that contains that information. These are kind of useful, when scanning to figure out when a particular MAF output was generated (users looking at the showmaf pages have no other way of getting this information), so it would be useful to generate this and put it somewhere (even if it's just by adding another table to the resultsDb which contains "maf was run on X date and with Y version" in a row, and then potentially adding another row when new outputs were added .. note that Eric and other people may use summary results db files a bit differently than we typically do). This is the reason why I requested not to remove the writeConfig function entirely, even if it's just limited to writing a summary config file with the maf information. 

Screen Shot 2022-01-05 at 2 37 33 PM

rhiannonlynne:
And then regarding labelling visits -- this is currently the easiest way to identify which parts of the sky are part of the WFD. So, I agree this is not the best way to do it but we haven't made it easy to do otherwise, given that recreating the original footprint (in order to pull out WFD regions) requires (I think) looking at the script which was used to generate the particular opsim run .. which I don't think is entirely straight forward to get in the first place, and then pulling the footprint in particular out of the script is a secondary thing that would require some work by hand. 
While that work is in progress, it's not obvious to me why the function needed to be deleted in this PR?

yoachim:
OK, MAF version and date get written to the `resultsDb_sqlite.db` and passed to `showMaf`. 

The labeling script called database methods that no longer exist, so it's broken. My plan is to update the schema to remove `propId` and `fieldId` completely, so any patch would be short-lived.

rhiannonlynne (approved):

PR #109, opened 2021-12-21
Upgraded TDEsPop metric to evaluate lightcurve quality

Bricmank:
Added a new class called the "TdePopMetricQuality" to the original TDEsPopMetric.py file, with which each light curve that  passed the "some_color" or "some_color_pu" original metric is assigned a score, based on how well sample it is. If the TDE does not pass some_color/some_color_pu requirements, the score is 0, if it is observed every other night, the score is 1, if it is observed every 4 nights, the score is 0.5. The score scales linearly with the number of observations in a predefined time interval, in which we wish to see how the light curve is sampled. This time interval is by default between -30 d < t_peak < 100 d.


rhiannonlynne:
From TVS meeting 1/17 - 
* run the new metric as well as the old metric
* what summary statistics should be used? -- median value is the primary goal, but run all the standard summary stats as well
-- will eventually have to be balanced with the total number of TDE from the original metric. This one reports on "how good is the coverage for a given TDE". 

Need further feedback regarding how to balance these (is a 5% increase in the total number of TDE worth a 5% decrease in coverage? or vice versa? or what ratios?)

PR #110, opened 2021-12-31
Added microlensingFOM, microlensingMetric Fisher, and updated microlensingMetric base magnitudes

nsabrams:
This PR has four components:

### Updated base magnitudes in microlensingMetric.py
The baseline magnitudes in microlensingMetric were updated from being a magnitude of 20 across all bands to being the median magnitude of each flter from the TRIstarDensity map:
- u: 25.2
- g: 25.0
- r: 24.5
- i: 23.4
- z: 22.8
- y: 22.5

Note that these are all dimmer than 20 mag, so the results of the microlensingMetric will be less efficient than previously.

### Added microlensingMetric Fisher
To characterize microlensing events, we calculate the Fisher information matrix from which we can determine the uncertainty in parameters. Here we return sigma_tE/tE where tE is the Einstein crossing time of the event. Note that calculating this matrix is more computationally expensive than the microlensing detect or microlensing Npts metrics.

A blend ratio of 0.5 (i.e. source flux = blend flux) is assumed for all objects in this method, since when they are drawn from N^2, they are likely in the most dense areas, so this is a reasonable estimate.
Contact Markus Hundertmark (markus.hundertmark@uni-heidelberg.de) for further details.

### Added microlensingFOM.py
Code to produce plots to compare between OpSims was written. The format of the figure is one subplot for the detect/discovery metric, one for the Npts metric, and one for the Fisher metric. Each subplot has the OpSim on the y-axis and the discovery efficiency, average number of points (logscale), and characterization efficiency (fraction of events with sigma_tE/tE < 0.1), respectively, on the x-axis. It expects that all three of these were run on the same OpSims with the same tE bins. 

This script also contains code to produce confusion-matrix-like plots comparing the three microlensing metrics. Running these would be useful to the TVS microlensing subgroup to compare the characterization and detection efficiencies.  It compares the fraction of lightcurves discovered, the fraction characterized (sigma_tE/tE < 0.1), and the fraction with at least 10 points within 2tE of the peak. (Note that sigma_tE/tE < 0.1 and at least 10 points are both defaults which can be overwritten).

If provided with a folder of .npz files and .db files with the appropriate naming conventions is provided (where the .npz file is create with the default naming and the .db file is the same with _result at the end), then the plots will be created. The plotting functions can also be called individually.

### Machinery to integrate with example notebook
In order to use this code with the example notebook (see Microlensing Metric.ipynb in the rubin_sim_notebooks repository), microlensingSummary was added to rubin_sim.maf.batches.common. There are two sets of summaries depending on whether "Npts" or "Fisher" is specified (for "detect", lightcurveSummary should be used). For Fisher, a new metric in simpleMetrics was added: RealMeanMetric where the mean of a metric is calculated excluding nans and infs.


rhiannonlynne:
Hi @nsabrams - quick question, related to the update in the mean magnitudes for the background stars -- do these new magnitudes come from the median magnitudes of the TriLegal star maps out to a certain distance, or in a particular l/b location, and does it include dust extinction?
(I guess I am not quite understanding what the median magnitude from the TriLegal star maps means).

rhiannonlynne:
We'll have to think a bit about how to incorporate the "comparing opsims" script. It's very useful, but we haven't really dealt with figures of merit which are *only* plots previously. Most of the FoM comparisons can be summarized as a single number per run, which can then be fairly easily compared between runs. (I will add, I have written this without looking at the code .. perhaps your FoM is essentially a couple of (independent) numbers, which could then be tracked and compared within our standard approach).

rhiannonlynne:
One last question: nans and infs? Should those values be masked instead of returned with the metric results? This would make the RealMeanMetric unnecessary, and is generally the approach we take for other metric results. (i.e. if the metric would return a value which is NaN, it's a 'bad value' and then ends up masked).

nsabrams:
> Hi @nsabrams - quick question, related to the update in the mean magnitudes for the background stars -- do these new magnitudes come from the median magnitudes of the TriLegal star maps out to a certain distance, or in a particular l/b location, and does it include dust extinction? (I guess I am not quite understanding what the median magnitude from the TriLegal star maps means).

I took the TriLegal CDF, turned it into a PDF, and took the overall mean for each filter. Would you like me to send over the code I used to do so? Does the map also include very distant stars outside the Milky Way, meaning that we should specify, all stars out to 8 or 15 kpc or something?

> We'll have to think a bit about how to incorporate the "comparing opsims" script. It's very useful, but we haven't really dealt with figures of merit which are _only_ plots previously. Most of the FoM comparisons can be summarized as a single number per run, which can then be fairly easily compared between runs. (I will add, I have written this without looking at the code .. perhaps your FoM is essentially a couple of (independent) numbers, which could then be tracked and compared within our standard approach).

Yes, the FoM is essentially 3 independent numbers generated for the detect, Npts, and Fisher metrics respectively. Those are the ones from the bundle summaries in maf.batches.common:

- For detect it's: lightcurveSummary - "Total detected"
- For Npts it's: microlensingSummary with metricType = 'Npts' - "Mean number of points per lightcurves in total"
- For Fisher it's: microlensingSummary with metricType = 'Fisher' - "Fraction w/ sigma_tE/tE < 0.1"

> One last question: nans and infs? Should those values be masked instead of returned with the metric results? This would make the RealMeanMetric unnecessary, and is generally the approach we take for other metric results. (i.e. if the metric would return a value which is NaN, it's a 'bad value' and then ends up masked).

Ah okay, sure I'll edit that. Is there an example of how this is done that I can follow?

rhiannonlynne:
To return a value that will end up masked, just return the metric "self.badval"  -- so you could test to see if you will get a nan or inf, or you could test to see if your calculated value is nan or inf .. and then return self.badval from the Metric instead.

yoachim:
Moving this over to PR #123

PR #113, opened 2022-01-04
U/yoachim/unit test path

yoachim:
Make it so unit tests are independent of what directory they are run in. Handy when one is working in just one sub-package and you don't want to run all the unit tests.


yoachim:
ugh, whatever. Fine, you need to run unit tests from the top level.

PR #115, opened 2022-01-05
U/yoachim/surfacebmetric

yoachim:
A simple surface brightness limit metric.

### DIFF
--- a/rubin_sim/maf/metrics/surfbMetric.py
+++ b/rubin_sim/maf/metrics/surfbMetric.py
@@ -4,38 +4,51 @@ from rubin_sim.photUtils import Bandpass, PhotometricParameters
 import rubin_sim.utils as rsUtils
 from rubin_sim.data import get_data_dir
 import os
+import warnings
 
 __all__ = ["SurfaceBrightLimitMetric"]
 
 
 def surface_brightness_limit_approx(
-    zp, k, airmass, mu_sky, rn=8.8, pixscale=0.2, nsigma=3.0, t_exp=30.0, tot_area=100.0
+    zp,
+    k,
+    airmass,
+    mu_sky,
+    rn=8.8,
+    pixscale=0.2,
+    nsigma=3.0,
+    t_exp=30.0,
+    tot_area=100.0,
+    mag_diff_warn=0.1,
 ):
     """Compute surface brightness limit in 3 limiting cases, return the brightest.
 
-    Algerbra worked out in this notebook: 
-    https://github.com/yoachim/21_Scratch/blob/main/surfb_metric/surface_brightness_limit.ipynb
+    Algerbra worked out in this technote:
+    https://github.com/lsst-sims/smtn-016
 
     Parameters
     ---------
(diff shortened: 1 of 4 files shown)


rubin_sim/maf/metrics/surfbMetric.py
```diff
@@ -0,0 +1,180 @@
...
+__all__ = ["SurfaceBrightLimitMetric"]
+
+
+def surface_brightness_limit_approx(
+    zp, k, airmass, mu_sky, rn=8.8, pixscale=0.2, nsigma=3.0, t_exp=30.0, tot_area=100.0
+):
+    """Compute surface brightness limit in 3 limiting cases, return the brightest.
+
+    Algerbra worked out in this notebook: 
+    https://github.com/yoachim/21_Scratch/blob/main/surfb_metric/surface_brightness_limit.ipynb
```
ehneilsen, line 17: The derivation should probably go in a note, or at least a repo outside of your personal github account.

I have not checked the math, except to notice that the "combining limiting magnitudes" section is a re-derivation of the definition of and standard use-case for t_eff.

rubin_sim/maf/metrics/surfbMetric.py
```diff
@@ -0,0 +1,180 @@
...
+        - k * airmass
+    )
+
+    # Readnoise limited case
+    mu_rn_lim = (
+        -2.5 * np.log10(nsigma * rn / (t_exp * A_pix * n_pix ** 0.5)) + zp - k * airmass
+    )
+
+    result = np.vstack([mu_sky_lim, mu_source_lim, mu_rn_lim])
+    result = np.min(result, axis=0)
```
ehneilsen, line 73: This might be off if no single origin for the noise dominates, e.g. if the readnoise is about the same as the sky noise and source noise is negligable, this will underestimate the noise by a factor of sqrt(2), so overestimate the depth by 2.5*log10(sqrt(2))=0.38, or more if all 3 are comparable.

ehneilsen (commented):
Mostly looks good. Need to figure out we can live with it being off by a few tenth of a mag in rare circumstances (when no one source on noise dominates).

rubin_sim/maf/metrics/surfbMetric.py
yoachim: Good idea, tech note made. I think I broke the sims tech notes by abandoning one a few months ago, so I'll ping the docs folks so it actually gets published.

rubin_sim/maf/metrics/surfbMetric.py
yoachim: Yup. Added a check so it'll throw a warning if it's doing a potentially lousy approximation.

ehneilsen: Good enough, unless we later find the warning gets thrown regularly. If we find that it is, we may need to combine them properly, but that can be a separate issue, not part of this commit.

ehneilsen (approved):

PR #123, opened 2022-01-10
U/yoachim/microlensing update

yoachim:
moving other PR to a branch here so it's easy to edit


yoachim:
Hi @nsabrams, I merged your work onto a branch so I could make a few minor edits and formatting things.

Two questions--

1) We don't currently have sympy on our dependency list. It seems to work if I just comment that part out. Is there a significant performance boost if sympy is used? If so I'll add it to the `requirements.txt`, otherwise I'm tempted to leave it out.

2) I see you made `microlensingSummary`. Our default set of metrics we run on new simulations is in the `scienceRadarBatch.py`. Should we update that to run the microlensing in Fisher mode, or leave as is?

yoachim:
On your `plot_FOM`, it might be easier if the mean tE was on the x-axis, metric value on the y-axis, and the run name in the legend.

rubin_sim/maf/mafContrib/microlensingFOM.py
```diff
@@ -0,0 +1,550 @@
+import numpy as np
+import matplotlib.pyplot as plt
+import warnings
+
+import glob
+import os
+import pandas as pd
+import rubin_sim.maf.db as db
+import rubin_sim.maf.metricBundles as metricBundles
+
```
rhiannonlynne, line 10: I think most of the contents of this file would find a better home under 'RunComparison' rather than mafContrib. Down the road, we're not expecting to find plotting methods in the mafContrib directory and are likely to overlook them .. likewise, comparison of outputs between different runs is a better fit for the RunComparison directory where we will be more likely to find them. 
Maybe just rename the file microlensingComparison.py and then move it to maf/runComparison/microlensingComparison.py?

nsabrams: > Hi @nsabrams, I merged your work onto a branch so I could make a few minor edits and formatting things.
> 
> Two questions--
> 
> 1. We don't currently have sympy on our dependency list. It seems to work if I just comment that part out. Is there a significant performance boost if sympy is used? If so I'll add it to the `requirements.txt`, otherwise I'm tempted to leave it out.
> 2. I see you made `microlensingSummary`. Our default set of metrics we run on new simulations is in the `scienceRadarBatch.py`. Should we update that to run the microlensing in Fisher mode, or leave as is?

1. We just found the optimization of those equations with sympy outside the metric. You don't actually need sympy to run the code.
2. Is it possible for you guys to run all three: detect, Npts, and Fisher? The latter two are quite helpful for determining how well we can characterize the events, instead of just seeing that there is a microlensing event. 


> On your `plot_FOM`, it might be easier if the mean tE was on the x-axis, metric value on the y-axis, and the run name in the legend.

I have it like that since if you have a very long list of metrics it can get pretty overwhelming. I can put the code in to do either way though if that'd be helpful? Here's an example of that kind of plot.

image


> I think most of the contents of this file would find a better home under 'RunComparison' rather than mafContrib. Down the road, we're not expecting to find plotting methods in the mafContrib directory and are likely to overlook them .. likewise, comparison of outputs between different runs is a better fit for the RunComparison directory where we will be more likely to find them. Maybe just rename the file microlensingComparison.py and then move it to maf/runComparison/microlensingComparison.py?

Sure sounds good. I just put it there since I thought all community contributions went to that folder.

rhiannonlynne:
One request for @yoachim  - could you move the microlensing metrics into their own batch? That way, it's a bit easier if we want to add the long-running microlensing metric later I think. We could just run the microlensing batch on top of the other metrics (updating the resultsDb, etc.), instead of having to hack the whole ScienceRadarBatch batch.

yoachim:
> One request for @yoachim - could you move the microlensing metrics into their own batch? That way, it's a bit easier if we want to add the long-running microlensing metric later I think. We could just run the microlensing batch on top of the other metrics (updating the resultsDb, etc.), instead of having to hack the whole ScienceRadarBatch batch.

I would advocate going the other way and say the `ScienceRadarBatch` should not be calling additional batch generators because that makes it difficult to find the code that was actually run to generate plots and we get messy re-organizing of displayDicts. 

Actually, just adding lots of flags all over the place to run/not-run various sections would solve the problem as well. I'll make a ticket to do that later.

PR #125, opened 2022-01-11
U/lynnej/galactic plane metrics

rhiannonlynne:
Currently work in progress. 
@rachel3834 tagging you here so you can see the tweaks.

Basically, I want to change the imports to be consistent with other rubin_sim imports, combine some of the galactic metrics into one file for easier location/editing later, have to add the map to the data repository (and update the version here to match). 
I'd also like to check in on the TauObsData object -- could it just be a dictionary or is there a particular reason it's an object? It doesn't look like the information is shared between instances of the object, so it seems kind of like it's only used to hold the information on what the timescales are? (like, if I made a constant array of the timescales, it would work in the same way?)


rachel3834:
Hi Lynne,

Thanks for your feedback.  Yes, I'm using the TauObsData object to store
the 4 values that are calculated for this metric.  Previously I was
handling these as a list (a dictionary would be more self-documenting
perhaps) but they could equally be an array.  I know we previously
discussed how to approach metrics that need to return more than one value
and I had understood that this was the preferred method, but I'm happy to
change it if you prefer.  Let me know what works best.

rachel3834:
This PR can be disregarded as it has been superceded by a replacement PR.

rhiannonlynne:
> This PR can be disregarded as it has been superceded by a replacement PR.

We cleared this up - this PR is from the branch to main and should *not* be disregarded. :)

rhiannonlynne:
I talked with Rachel, cleared a few things up about how these metrics are working.
She's going to refactor, on a new branch at her end, and reissue a PR. 
I'll close this PR accordingly (and will delete this branch, to help us clean up work between the two forks.).

PR #133, opened 2022-01-14
prevent specifying alternate names for metrics from clobbering style …

ehneilsen:
Using different metric labels could confuse the old code, because the style information was still associated with the old "official" metric names. Code on this branch updates the index used to find the style at the same time it updates the names, preserving the mapping between metric and style.


rhiannonlynne:
Tested it out - looks like this works! :)

rhiannonlynne (approved):

PR #134, opened 2022-01-18
fix typo

yoachim:
That makes a difference!


ehneilsen (approved):
Yes, that looks better.

PR #140, opened 2022-01-21
minor updates

yoachim:
Add XRB data file. Add L4 brown dwarfs to science radar.

### DIFF
--- a/rubin_sim/maf/batches/scienceRadarBatch.py
+++ b/rubin_sim/maf/batches/scienceRadarBatch.py
@@ -504,17 +504,28 @@ def scienceRadarBatch(
     bundleList.extend(temp_list)
 
     # XRB metric
-    n_events = 10000 
+    displayDict["subgroup"] = "XRB"
+    n_events = 10000
     slicer = generateXRBPopSlicer(n_events=n_events)
     metric = XRBPopMetric(outputLc=True)
-    xrb_summaryMetrics = [maf.SumMetric(metricName='Total detected'), 
-                      maf.CountMetric(metricName='Total lightcurves in footprint'),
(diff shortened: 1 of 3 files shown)


rubin_sim/maf/batches/scienceRadarBatch.py
```diff
@@ -503,6 +503,30 @@ def scienceRadarBatch(
         temp_list.append(bundles[b])
     bundleList.extend(temp_list)
 
+    # XRB metric
+    displayDict["subgroup"] = "XRB"
+    n_events = 10000
+    slicer = generateXRBPopSlicer(n_events=n_events)
+    metric = XRBPopMetric(outputLc=True)
+    xrb_summaryMetrics = [
```
rhiannonlynne, line 511: Isn't this set just the "lightcurveSummary"? https://github.com/lsst/rubin_sim/blob/main/rubin_sim/maf/batches/common.py#L151

rhiannonlynne: That said, I think I'd add one more metric on top of these that is just the "MedianMetric" and just call it the median .. that would likely be useful for the reduced "Number of detections" and "Time between detections" and it would also have a straight forward name (it'd be useless for the ever detect / early detect, but for the other two would be ok and clearer)

rubin_sim/maf/batches/scienceRadarBatch.py
```diff
@@ -596,6 +620,21 @@ def scienceRadarBatch(
         )
     )
 
+    metric = metrics.BDParallaxMetric(
```
rhiannonlynne, line 623: What is this reporting again? is it just the number of brown dwarfs that have parallax error less than a given amount?

rhiannonlynne (approved):

PR #144, opened 2022-01-27
U/yoachim/example fixes

yoachim:
* Some docstring updates. 
* Make the rotation angle a little more sophisticated so we can set a desired rotSkyPos and the observatory does reasonable things if it's not available.

### DIFF
--- a/rubin_sim/scheduler/modelObservatory/model_observatory.py
+++ b/rubin_sim/scheduler/modelObservatory/model_observatory.py
@@ -450,6 +450,11 @@ class Model_observatory(object):
 
     def _update_rotSkyPos(self, observation):
         """If we have an undefined rotSkyPos, try to fill it out."""
+
+        # Grab the rotator limit from the observatory model
+        rot_limit = [self.observatory.telrot_minpos_rad+2.*np.pi,
+                     self.observatory.telrot_maxpos_rad]
+
         alt, az = _approx_RaDec2AltAz(
             observation["RA"],
             observation["dec"],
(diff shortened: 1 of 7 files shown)


rubin_sim/scheduler/modelObservatory/model_observatory.py
```diff
@@ -450,16 +450,44 @@ def check_mjd(self, mjd, cloud_skip=20.0):
...
             self.site.longitude_rad,
             self.mjd,
         )
+
         obs_pa = _approx_altaz2pa(alt, az, self.site.latitude_rad)
-        observation["rotSkyPos"] = (obs_pa + observation["rotTelPos"]) % (2 * np.pi)
-        observation["rotTelPos"] = 0.0
+
+        # If the observation has a rotTelPos set, use it to compute rotSkyPos
+        if (np.isfinite(observation["rotSkyPos"])) | (
```
ehneilsen, line 471: why is np.isfinite(observation["rotSkyPos"]) evaluated twice here? isn't  "x|x" always "x"?

rubin_sim/scheduler/utils/utils.py
```diff
@@ -436,49 +440,41 @@ def opsim2obs(self, filename):
...
-    XXX-Could add a bool flag for "observed". Then easy to track all proposed
-    observations. Could also add an mjd_min, mjd_max for when an observation should be observed.
-    That way we could drop things into the queue for DD fields.
-
-    XXX--might be nice to add a generic "sched_note" str field, to record any metadata that
-    would be useful to the scheduler once it's observed. and/or observationID.
-
     Returns
     -------
     numpy array
```
ehneilsen, line 445: If you want to pedantically follow DM standards, the [return needs to be named](https://developer.lsst.io/python/numpydoc.html#py-docstring-returns).

ehneilsen (approved):
Comments are minor, looks good.

PR #146, opened 2022-01-28
U/yoachim/xyplotter

yoachim:
Way to make a simple viz of how the DDF observations progressed. Should be useful for checking that rolling and accordion cadences worked properly.

### DIFF
--- a/rubin_sim/maf/metrics/cumulativeMetric.py
+++ b/rubin_sim/maf/metrics/cumulativeMetric.py
@@ -11,9 +11,9 @@ class CumulativeMetric(BaseMetric):
 
     Parameters
     ----------
-    bins : `np.array` (None)
+    interp_points : `np.array` (None)
         The points to interpolate the cumulative number of observations to. If None,
-        then the range of the data is used.
+        then the range of the data is used with a stepsize of 1.
     """
 
     def __init__(
@@ -21,7 +21,7 @@ class CumulativeMetric(BaseMetric):
         metricName="Cumulative",
         time_col="observationStartMJD",
         night_col="night",
-        bins=None,
+        interp_points=None,
         **kwargs
     ):
         super().__init__(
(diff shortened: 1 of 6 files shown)


rubin_sim/maf/metrics/cumulativeMetric.py
```diff
@@ -0,0 +1,47 @@
...
+            **kwargs
+        )
+        self.time_col = time_col
+        self.night_col = night_col
+        self.bins = bins
+        self.plotDict = {"xlabel": "MJD (days)", "ylabel": "N obs"}
+
+    def run(self, dataSlice, slicePoint=None):
+        dataSlice.sort(order=self.time_col)
+        if self.bins is None:
```
rhiannonlynne, line 40: I don't understand why this works, so I must be missing something. 
If you're counting the number of visits in each night, I would have thought you could just use np.histogram. 
So then I tried both of these with a very stupid test case, and I don't get results that make sense to me. 
Screen Shot 2022-02-01 at 2 14 23 PM

rubin_sim/maf/plots/xyplotter.py
```diff
@@ -0,0 +1,31 @@
...
+
+        fig = plt.figure(fignum)
+        ax = fig.add_subplot(111)
+        x = metricValueIn[0]["x"]
+        y = metricValueIn[0]["y"]
+        ax.plot(x, y)
+        ax.set_title(plotDict["title"])
+        ax.set_xlabel(plotDict["xlabel"])
+        ax.set_ylabel(plotDict["ylabel"])
+        return fig.number
```
rhiannonlynne, line 31: This doesn't seem unreasonable, but you could also make "x" and "y" user-configurable to plot any two metric values in a dictionary or recarray against each other.

rhiannonlynne (commented):
Comments in code.

rhiannonlynne:
Sorry for the delay. I guess covid can definitely vary widely, but still getting over it here.

rubin_sim/maf/plots/xyplotter.py
yoachim: A plotter only gets one metric value at a time though, so I'm not sure how/why you would plot different metric values against each other.

rubin_sim/maf/metrics/cumulativeMetric.py
yoachim: Ah, `bins` is a terrible kwarg name. There's no actual binning happening. I just put the interpolation step in there to avoid plotting 2 million line segments. Updated to a better name.

rubin_sim/maf/plots/xyplotter.py
rhiannonlynne: I wondered if my comment would sound like that and yes - I get that, but what I meant was other metrics which are returning more than one value (most of which have reduce functions) .. but could find it useful to plot their various values against each other.  Obviously nobody's doing that now, but it's not unreasonable .. like if you calculated season length and season cadence in one metric returning them both, then you could see how they were correlated in one step. 
Will this plotter fail if there's more than one data point in metricValues?

rubin_sim/maf/metrics/cumulativeMetric.py
rhiannonlynne: Ok, I see. And the other thing I was missing was a sort.

rubin_sim/maf/plots/xyplotter.py
rhiannonlynne: Anyway - this is fine. I think if the plotter will only work if there's only one data point, then it's worth making a comment that this is so in the docstring, and making it configurable later is reasonable. 
(one note is that if you did make the x/y values user-configurable to be the names of what the metric returned, then those would also be reasonable to take as the labels for those axes).

rhiannonlynne (approved):

PR #149, opened 2022-02-04
U/lynnej/3d dustmap

rhiannonlynne:
Working 3d dust map. 
(todo: run YSO notebook to check usage)


yoachim:
Looks like this breaks `mafContrib.NYoungStarsMetric`.

rhiannonlynne:
> Looks like this breaks `mafContrib.NYoungStarsMetric`.

Indeed. I hit "request PR" a bit too soon, as I found out when I went to the notebook (I went back and unrequested PR, but I suppose I should have closed the whole PR instead). 
It works now.

yoachim (approved):
Looks good. Might be worth explicitly adding to the `DustMap3D` docstring what new keys are getting added to the slicePoint (and their units and all that).

PR #150, opened 2022-02-07
U/yoachim/small pre

yoachim:
Add a 8-day pre-computed sky brightness, so folks can run scheduler examples and short tests without needing many GBs of data.

### DIFF
--- a/.github/workflows/python-tests-doc.yml
+++ b/.github/workflows/python-tests-doc.yml
@@ -48,7 +48,7 @@ jobs:
         shell: bash -l {0}
         run: |
           export RUBIN_SIM_DATA_DIR=${{ github.workspace }}/data_dir
-          rs_download_data -d 'site_models,throughputs,skybrightness,skybrightness_pre'
+          rs_download_data -d 'site_models,throughputs,skybrightness,skybrightness_pre,skybrightness_pre_small'
           rs_download_data -d tests --force
       - name: conda list
         shell: bash -l {0}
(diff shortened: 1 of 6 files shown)


rhiannonlynne:
Oh hey - would you want to have a look at 
https://github.com/lsst/rubin_sim/blob/0fbe0c398b504db12d9ecd9266b4f2c119050247/rubin_sim/skybrightness_pre/data/generate_sky.py
and just check if that version creates the same / equivalent skybrightness files as here? (thus dropping ephem and some of the output columns that are unneeded in the skybrightness pre file)?

yoachim:
Yup, that looks like it generates the same data.

rhiannonlynne:
Why not put the small skybrightness pre file into the tests directory instead of making a new one?

yoachim:
> 

I guess because it's "real" in the sense that it's a file that can be used in-production. It's also 240M, and the tests dir is still only 41M and it seems like it might be nice to keep that smaller. Also, I expect this is a file that we'll want to clobber as the survey start-date moves around.

rhiannonlynne:
Seems reasonable. 
Unit tests still failing? (looks like two black errors .. and maybe the other ones are due to needing to add skybrightness_pre_small to the test download data for the GitHub actions which I'll go do now)   
But otherwise, is good.

rhiannonlynne:
To add skybrightness_pre_small to the downloaded data for the GitHub unit tests, can you please modify this line: 
`rs_download_data -d 'site_models,throughputs,skybrightness,skybrightness_pre'`
(line 51) in the .github/workflows)/python-tests-doc.yml file? 
(Or I can modify this and add a commit here, but you just need to change it to 
`rs_download_data -d 'site_models,throughputs,skybrightness,skybrightness_pre,skybrightness_pre_small'`

rhiannonlynne (approved):
sorry - didn't realize I hadn't actually hit the "approve" button

PR #153, opened 2022-02-21
U/yoachim/ddf fiddle

yoachim:
* Expand the DDF batch
* Consolidate science radar so magic numbers are not scattered in many different files
* Update SNe utils to cache loaded data


rubin_sim/maf/mafContrib/__init__.py
```diff
@@ -4,6 +4,7 @@
 from .varMetrics import *
 from .varDepthMetric import *
 from .lssMetrics import *
+from .lssmetrics import *
```
rhiannonlynne, line 7: Wow, that is confusing because they look so alike. Can we just lift the single file out of the lssmetrics directory and put it in here?

rubin_sim/maf/utils/snNSNUtils.py
```diff
@@ -1038,6 +1039,19 @@ def gamma(self, mag, band, exptime):
...
+    doesn't need to load it on it's own
+    """
+
+    if hasattr(load_sne_cached, "data"):
+        return load_sne_cached.data
+    else:
+        ref = Load_Reference().ref
+        load_sne_cached.data = ref
+        return ref
+
```
rhiannonlynne, line 1053: Great idea.

rubin_sim/scheduler/surveys/scripted_surveys.py
```diff
@@ -205,6 +205,9 @@ def generate_observations_rough(self, conditions):
         observations = self._check_list(conditions)
         if observations is not None:
             observations = [self._slice2obs(obs) for obs in observations]
+        import pdb
+
+        pdb.set_trace()
```
rhiannonlynne, line 210: Should these pdb  statements be removed?

rhiannonlynne:
I thought that we had decided to go the other way on whether to create a giant file with a single science radar batch, versus having smaller modules which could be combined together in the science radar batch? 
I guess I don't care now it's all done. However, I am worried this wil be a bit hard to maintain. It's already a single file of code that is over 1400 lines long. Choosing where to insert new metrics so that they make sense and are easier to find later is going to be hard. 
I suggest that, if this is the approach we're going to settle on, that there need to be bigger and easier to see comment lines as you scan up and down, so that metrics for a given science case are easier to group together. 
(imagine trying to add more periodic star metrics, for example .. so having to find the section of the science radar code that deals with periodic stars, to add them there, and scanning up and down, potentially not having the right/same comment line to search for to get to the right piece of code..).

rhiannonlynne:
I'm somewhat unsure whether I would then also say you should remove
*agnBatch
*descWFDBatch
*periodicStarBatch
*srdBatch

I think you should, so they don't get out of sync (I could imagine someone only remembering to update the configurations in one or the other place, and then not remembering that the other one was out of date). 
But then I think you need to actually put in all of the flags in scienceRadar so you can run subsets of these if needed (particularly the srdBatch).

rhiannonlynne (approved):
Please at least add flags so we can run the SRD metrics without having to run all of the science metrics.

PR #155, opened 2022-02-24
U/rachel/galactic plane metrics

rachel3834:
I've revised the galPlaneFootprintMetric and galPlaneTimePerFilterMetric.py as discussed with Lynne.  Both have been dramatically simplified to calculate each metric for a single HEALpixel, assuming that a single science map (from the Galactic Plane priority map files) is passed as a kwarg at a time.  

The priority map files themselves have been updated to provide pre-calculated maps where the priority of each HEALpixel to the survey is summed over all filters.  These data are used in the revised metrics to minimize the computational overhead.


rhiannonlynne:
Hi @rachel3834 -- were you going to take the RGESOverlap metric out of this PR and submit it later? 
(it won't import because it's using a package we're not making available)

rachel3834:
Apologies that one was supposed to be on a different branch.  I've removed it.

rhiannonlynne:
Allrighty - thanks! Next request: please remove the unused imports (the gal footprint metric, for example, is still importing astropy_healpix).

rhiannonlynne:
OH wait - do you mean that we should *only* configure and run the GalPlaneFoM metric and then that one will call the other metrics and run those? And so we can run summary metrics on the GalPlaneFoM metric? 
In the GalPlaneFoM metric, you should init and assign to self.galPlaneFilterMetric (etc) each of these metrics in the Init of the GalPlaneFoM metric. (that way they only get set up and read the map once .. and then you can call their "run" methods with the dataSlice and slicePoint that the GalPlaneFoM metric receives).

rachel3834:
I've removed all references to astropy-healpix in the Galactic Plane metrics and I'm in the process of removing this package from the Roman/Rubin metrics as well, but that's on a different branch, so I've removed it from this one.  
I've also verified that my (separate) priority map generation code produces identical results when I switched that to use pure healpy rather than astropy-healpix as well. 

Having had more of a chance to analyse the metric results, I am of two minds about the GalacticFoM code.  It was really intended as an ultra-high-level guide for the SCOC but I fear that it could be so generalized as to be potentially misleading.  It might be best to compute the three separate metrics and evaluate the results.

PR #156, opened 2022-02-24
add warning for twilight extrapolation

yoachim:
Updates for folks on 4MOST who want to estimate sky values to very bright twilight times. Throw a warning if we are extrapolating values beyond the fitted range. Throw NaNs if trying to compute sky brightness with the sun high in the sky.

### DIFF
--- a/rubin_sim/skybrightness/interpComponents.py
+++ b/rubin_sim/skybrightness/interpComponents.py
@@ -545,7 +545,7 @@ class TwilightInterp(object):
         self,
         interpPoints,
         maxAM=3.0,
-        limits=[np.radians(15.0), np.radians(-20.0)],
+        limits=(np.radians(15.0), np.radians(-20.0)),
         filterNames=["u", "g", "r", "i", "z", "y"],
     ):
         """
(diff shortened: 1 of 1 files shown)


rubin_sim/skybrightness/interpComponents.py
```diff
@@ -562,15 +567,19 @@ def interpMag(
...
+                result[good, i] = twilightFunc(
+                    interpPoints[good],
+                    *self.lsstEquations[self.filterNameDict[filterName], :].tolist()
+                )
 
         return {"spec": result, "wave": self.lsstEffWave}
 
     def interpSpec(
-        self, interpPoints, maxAM=3.0, limits=[np.radians(-5.0), np.radians(-20.0)]
+        self, interpPoints, maxAM=3.0, limits=[np.radians(15.0), np.radians(-20.0)]
```
ehneilsen, line 582: It would be better for the default of "limits" to be a tuple (which is immutable) rather than a list (which is mutable, so can lead to unexpected consequences).

ehneilsen (approved):
Looks fine.

PR #157, opened 2022-02-24
U/lynnej/add comets

rhiannonlynne:
Adds OCC populations to solar system populations.
(occ_rmax5 and occ_rmax20)
Moves many of the functions from makeLSSTobs into rubin_sim.movingObjects.utils
Updates the generate_ss script to generate independent run-scripts per opsim; these could be more easily joined together, or alternatively, run in parallel without interference (the single ss_script.sh is gone). 
Example batch command for the new Oort Cloud occ populations:
```
for rr in `cat filelistX0`; do
  generate_ss --pop occ_rmax5_5k --db /gscratch/scrubbed/lynnej/fbs_2.0/fbs_2.0_db/$rr".db" --split
  source $rr"_ss_script.sh"
  generate_ss --pop occ_rmax20_5k --db /gscratch/scrubbed/lynnej/fbs_2.0/fbs_2.0_db/$rr".db" --split
  source $rr"_ss_script.sh"
  done
  ```
(should run with or without the --split, but I've mostly run it with)

### DIFF
--- a/bin/maf/generate_ss
+++ b/bin/maf/generate_ss
@@ -60,7 +60,6 @@ def generate_ss_commands(dbfiles=None, pops=None, start_mjd=60218.0, split=False
             os.mkdir(outDir)
         except FileExistsError:
             pass
-        # ss_script (and other run scripts) will be created in outDir
         outfile = f"{run}_ss_script.sh"
         with open(outfile, "w") as f:
             for pop in pops:
--- a/bin/maf/generate_ss
+++ b/bin/maf/generate_ss
@@ -56,22 +56,24 @@ def generate_ss_commands(
 
     runs = [os.path.split(file)[-1].replace(".db", "") for file in dbfiles]
     runs = [run for run in runs if "tracking" not in run]
-    with open(outfile, "w") as f:
-        for pop in pops:
-            objtype = objtypes[pop]
-            for run, filename in zip(runs, dbfiles):
+    for run, filename in zip(runs, dbfiles):
+        outDir = f"{run}_ss"
+        try:
+            os.mkdir(outDir)
+        except FileExistsError:
+            pass
+        # ss_script (and other run scripts) will be created in outDir
+        with open(outfile, "w") as f:
+            for pop in pops:
+                objtype = objtypes[pop]
                 if split:
                     splitfiles = glob.glob(
                         os.path.join(data_dir, "split") + f"/*{pop}*"
                     )
-                    outDir = f"{run}_ss"
                     outfile_split = outfile.replace(".sh", f"_{pop}_split.sh")
-                    try:
-                        os.mkdir(outDir)
-                    except FileExistsError:
-                        # Directory already exists .. maybe need to remove split file
-                        if os.path.isfile(outfile_split):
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: rubin_sim/maf/metricBundles/moMetricBundle.py [hunk 1]
```diff
@@ -86,6 +86,8 @@ def makeCompletenessBundle(bundle, completenessMetric, Hmark=None, resultsDb=Non
         mb.setSummaryMetrics(metric)
         mb.computeSummaryStats(resultsDb)
         val = mb.summaryValues["Value At H=%.1f" % Hmark]
+        if val is None:
```
yoachim, line 89: I had to add this here to get things to run. I think it's just a case where running on a very small example survey there can be no detections of a population, and thus the summaryValue ends up as a `None` which will fail later when it expects a float. But it would be good to double check this isn't actually a bug upstream.

yoachim (approved):
I had to fiddle some things to get it to run. Pretty sure it's working now, but it would be good to run a double check that it is still outputting everything it's supposed to.

bin/maf/generate_ss
```diff
@@ -119,6 +121,7 @@ def generate_ss_commands(dbfiles=None, pops=None, start_mjd=60218.0, split=False
                     s3 = (
                         f"run_moving_fractions --workDir {run}_ss"
                         f" --metadata {objtype}"
+                        f" --hMark {default_plots_values['Hmark']} "
```
rhiannonlynne: Hmm, I thought the hMark should have been recorded in the metric file when it was configured at step1 (from the movingObjectBatch), so wouldn't have had to be repeated here. But it's also not hurtful. 
Worth noting - the values returned from ss_population_defaults are not just for the plots. 
Your renaming of the function is much improved, but I'm going to update the returned values to indicate the 'not just plots' part as well.

bin/maf/generate_ss
```diff
@@ -54,74 +54,80 @@ def generate_ss_commands(dbfiles=None, pops=None, start_mjd=60218.0, split=False
 
     runs = [os.path.split(file)[-1].replace(".db", "") for file in dbfiles]
     runs = [run for run in runs if "tracking" not in run]
+    if not split:
+        output_file = open("ss_script.sh", "w")
```
rhiannonlynne, line 58: I wasn't sure if it was helpful for your use to have a single file or not. It sounded like you probably would have found multiple names more useful, so you could run in parallel. 
If this is fine, sounds good.

rubin_sim/maf/metricBundles/moMetricBundle.py
rhiannonlynne: Can you provide an example case where it fails? I don't think I've seen a problem, so it may indeed be only for odd cases with no detections. Reasonable to catch it though.

bin/maf/generate_ss
rhiannonlynne: (or maybe I won't bother .. it looks like it is only being used for the plots here)

rhiannonlynne:
Ran through and tested it again; looks like everything works ok.

PR #158, opened 2022-02-25
U/yoachim/skysaturation

yoachim:
Updates to saturation stacker, new sky saturation metric, patches to glance batch.

### DIFF
--- a/rubin_sim/maf/batches/glanceBatch.py
+++ b/rubin_sim/maf/batches/glanceBatch.py
@@ -342,6 +342,94 @@ def glanceBatch(
         metricb.summaryMetrics = []
         bundleList.append(metricb)
 
+    # Add a sky saturation check
+    displayDict = {}
+    displayDict["group"] = "Basic Stats"
+    displayDict["subgroup"] = "Saturation"
+    sql = ""
+    metric = metrics.SkySaturationMetric()
+    summary = metrics.SumMetric()
+    slicer = slicers.UniSlicer()
+    bundleList.append(
+        metricBundles.MetricBundle(
+            metric, slicer, sql, summaryMetrics=summary, displayDict=displayDict
+        )
+    )
+
+    benchmarkArea = 18000
+    benchmarkNvisits = 825
+    minNvisits = 750
+    displayDict = {"group": "SRD", "subgroup": "FO metrics", "order": 0}
+
+    # Configure the count metric which is what is used for f0 slicer.
+    metric = metrics.CountExplimMetric(col="observationStartMJD", metricName="fO")
+    plotDict = {
+        "xlabel": "Number of Visits",
+        "Asky": benchmarkArea,
+        "Nvisit": benchmarkNvisits,
+        "xMin": 0,
+        "xMax": 1500,
+    }
+    summaryMetrics = [
+        metrics.fOArea(
+            nside=nside,
+            norm=False,
+            metricName="fOArea",
+            Asky=benchmarkArea,
+            Nvisit=benchmarkNvisits,
+        ),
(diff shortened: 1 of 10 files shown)


rubin_sim/maf/stackers/generalStackers.py
```diff
@@ -67,24 +70,31 @@ def __init__(
         self.filterCol = filterCol
         self.airmassCol = airmassCol
         self.saturation_adu = saturation_e / gain
-        self.pixscale = 0.2
-        names = ["u", "g", "r", "i", "z", "y"]
-        types = [float] * 6
+        self.pixscale = pixscale
+        self.saturation_adu = saturation_e
```
rhiannonlynne, line 74: It's confusing having this line with line 72 above. 
Replace line 72 with line 74 and add a comment about why not using saturation_adu? 
Or alternatively, may make sense to replace saturation_adu entirely, as it seems like the rest of the calculations will only work with e- and not adu, given how the zero points are being calculated.

rubin_sim/maf/batches/glanceBatch.py
```diff
@@ -342,6 +342,94 @@ def glanceBatch(
...
+            metric, slicer, sql, summaryMetrics=summary, displayDict=displayDict
+        )
+    )
+
+    benchmarkArea = 18000
+    benchmarkNvisits = 825
+    minNvisits = 750
+    displayDict = {"group": "SRD", "subgroup": "FO metrics", "order": 0}
+
+    # Configure the count metric which is what is used for f0 slicer.
```
rhiannonlynne, line 364: This seems like duplicate code to what is in the scienceRadarBatch?

rhiannonlynne (approved):

rhiannonlynne:
Note that the documentation failure here is a real failure, not an "upload to S3 failed" (which seems to be a temporary thing brought on by race conditions in building docs). 
Here the documentation failure is because of something with the new saturation metric being part of the glanceBatch .. which is surprising to me, so I must have forgotten something that has to happen to build the documentation.

rhiannonlynne:
We should make this snippet a utility somewhere: 
```        # Compute default zeropoints
        if zpt is None:
            zp_inst = {}
            datadir = get_data_dir()
            for filtername in "ugrizy":
                # set gain and exptime to 1 so the instrumental zeropoint will be in photoelectrons and per second
                phot_params = PhotometricParameters(
                    nexp=1, gain=1, exptime=1, bandpass=filtername
                )
                bp = Bandpass()
                bp.readThroughput(
                    os.path.join(
                        datadir, "throughputs/baseline/", "total_%s.dat" % filtername
                    )
                )
                zp_inst[filtername] = bp.calcZP_t(phot_params)
            self.zpt = zp_inst
        else:
            self.zpt = zpt
        if kAtm is None:
            syseng = rsUtils.SysEngVals()
            self.kAtm = syseng.kAtm
        else:
            self.kAtm = kAtm
            ```
           since I've seen it in at least two places (with slightly different self setups I suspect).

rhiannonlynne:
I added the throughputs to the data required for building documentation. I believe this should fix it. 
I am .. not entirely sure why it was triggered when adding the sky saturation metric to the glance batch however, but it was part of the auto mod documentation. 
I think it was something like this: 
the auto mod documentation could document the skysaturation metric (which uses the stacker column) without needing to run the stacker, so all was fine.
but when it tried to document the glanceBatch, for some reason it is actually running the init of the stacker, which requires reading the throughputs files. 
In the future, it is worth figuring out exactly why the stacker init was being run from the batch but not from the class; and also, considering putting "heavier" reads of things in a method which will not be triggered simply by building documentation.

PR #161, opened 2022-02-28
U/yoachim/move presto load

yoachim:
Change the presto KNe metric so the large loaded data blobs are not metric attributes (and thus do not get copied when running reduce methods).

### DIFF
--- a/rubin_sim/maf/mafContrib/prestoColorKNePopMetric.py
+++ b/rubin_sim/maf/mafContrib/prestoColorKNePopMetric.py
@@ -28,16 +28,18 @@ def radec2gal(ra, dec):
     return gal_l, gal_b
 
 
-def _load_hash(fileGalactic="TotalCubeNorm_1000Obj.pkl",
-               fileExtragalactic="TotalCubeNorm_1000Obj.pkl",
-               skyregion="extragalactic"):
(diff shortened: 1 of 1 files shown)


rubin_sim/maf/mafContrib/prestoColorKNePopMetric.py
```diff
@@ -28,6 +28,43 @@ def radec2gal(ra, dec):
...
 
+def _load_hash(
+    fileGalactic="TotalCubeNorm_1000Obj.pkl",
+    fileExtragalactic="TotalCubeNorm_1000Obj.pkl",
+    skyregion="extragalactic",
+):
+    """Helper function to load large hash table without being a metric attribute.
+    Note, bad things could happen if you try to run different sky regions at the same time
+    (like, it might thrash loading one then the other. So, keep that in mind if/when this
+    gets extended)
```
rhiannonlynne, line 39: Could you load them as two separate attributes of this function? like a dictionary .. and only load one when the sky region is 'activated'
(this seems similar to the ebvhp.py and ebv3Dhp.py and now galactic_plane_maps.py hack, so this seems like a good option)

rhiannonlynne (approved):

PR #165, opened 2022-03-03
make sure database connection is not kept open

yoachim:
Make it so ResultsDb object does not hold a connection open. This is a problem when running moving objects in parallel since different populations may try to write results to the same sqlite file at the same time. If the ResultsDb object is kept open, attempts to write from other threads timeout and fail.


rhiannonlynne:
I think earlier investigations showed that it wasn't a huge amount of overhead to open and close the sqlite database. 
Doing the open/close while writing data is probably useful, given the potential for clashes between different processes (I don't think this solves any potential clashes, but it reduces them greatly).

rhiannonlynne (approved):

PR #166, opened 2022-03-04
rearange and consolidate a bit

yoachim:
Eliminated some duplicate projections code. Moved useful functions up to utils so they are more generally available.


rhiannonlynne (approved):
Looks good, appreciate the added documentation.

PR #168, opened 2022-03-06
U/lynnej/fix tgaps slicer

rhiannonlynne:


yoachim:
If folks are demanding WFD-only numbers so much, should we just have a script that does a pass to figure out which HEALpix should be included in WFD, then run some set of metrics with the subset slicer?

yoachim (approved):

rhiannonlynne:
> If folks are demanding WFD-only numbers so much, should we just have a script that does a pass to figure out which HEALpix should be included in WFD, then run some set of metrics with the subset slicer?

Working on that now :)

PR #170, opened 2022-03-08
remove Error from names

yoachim:
Remove "Error" from metric names because it makes it hard to grep log files for actual errors.


ehneilsen (approved):
Looks reasonable to me.

PR #173, opened 2022-03-16
U/yoachim/skyrot mod

yoachim:
Quick fix so scheduler doesn't output negative RotTelPos values, and metric can deal with it even if it does.


ehneilsen (approved):
Looks fine.

PR #174, opened 2022-03-17
U/lynnej/change metadata

rhiannonlynne:
Changes "metadata" to "info_label" when used in MetricBundles. 
Also adds some methods to update resultsDbs on disk so that showMaf will work (the upgrade to 'metricInfoLabel' is handled automatically; downgrading to the old version with 'metricMetadata' has to be done by hand)
Adds a script that will run some basic all-sky and wfd-only analysis for things like airmass, limiting magnitude, seeing, inter night gaps, etc.


yoachim:
The `metadata_dir` fails if I run it on a short sim (~10 days). Looks like it's something in the plotting that's failing if it doesn't get any values.

bin/maf/metadata_dir
```diff
@@ -3,48 +3,18 @@
 import os
```
yoachim, line 3: `metadata_dir` needs to be added to the `setup.py` to get it in users' path.

bin/maf/metadata_dir
```diff
@@ -53,134 +23,157 @@ if __name__ == "__main__":
...
-            tags = ['All']
+        # Find the 'wfd' footprint
+        m = CountExplimMetric(col="observationStartMJD")
+        allsky_slicer = HealpixSlicer(nside=args.nside)
+        constraint = 'note not like "%DD%"'
+        bundle = MetricBundle(m, allsky_slicer, constraint, runName=opsim)
+        g = MetricBundleGroup({f"{opsim} footprint": bundle}, opsdb, outDir=outDir)
+        g.runAll()
+        wfd_footprint = bundle.metricValues.filled(0)
+        wfd_footprint = np.where(wfd_footprint > 750, 1, 0)
```
yoachim, line 79: magic number 750 should probably be an arg so it's documented as well.

rhiannonlynne: True!

bin/maf/metadata_dir
rhiannonlynne: Yeah, I know .. I also know you chose to leave it out when you set up the executables before, so I figured maybe there was a reason.

rhiannonlynne:
> The `metadata_dir` fails if I run it on a short sim (~10 days). Looks like it's something in the plotting that's failing if it doesn't get any values.

Is the short sim the same as the one in the rubin_sim_data/tests directory?

yoachim (approved):

PR #178, opened 2022-03-24
U/yoachim/observatory overhead

yoachim:
Refactor readtime as an overhead. Add shutter cool-down time an another possible contributor to overhead.

### DIFF
--- a/rubin_sim/scheduler/modelObservatory/kinem_model.py
+++ b/rubin_sim/scheduler/modelObservatory/kinem_model.py
@@ -104,6 +104,7 @@ class Kinem_model(object):
         maxspeed=3.5,
         accel=1.0,
         decel=1.0,
+        two_motion_time=15.0,
     ):
         """
         Parameters
@@ -124,6 +125,9 @@ class Kinem_model(object):
             The maximum speed of the rotator (degrees/s)
         accel : `float` (1.0)
             The acceleration of the rotator (degrees/s^2)
+        two_motion_time : `float` (15.)
+            The time required for two shutter motions (seconds). So, if one takes
+            a 1-snap 10s exposure, there will be a 5s overhead before the next exposure can start.
         """
         self.readtime = readtime
         self.shuttertime = shuttertime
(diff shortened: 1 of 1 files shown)


rhiannonlynne:
@smr456 Steve Ritz - we're adding the minimum 15s time between exposures here.

rubin_sim/scheduler/modelObservatory/kinem_model.py
```diff
@@ -605,19 +610,35 @@ def slew_times(
...
+
     def observe(self, observation, mjd, rotTelPos=None, lax_dome=True):
         """observe a target, and return the slewtime and visit time for the action
 
         If slew is not allowed, returns np.nan and does not update state.
         """
+        if (observation["nexp"] > 2) & (
+            observation["exptime"] / observation["nexp"] < self.two_motion_time
+        ):
+            warnings.warn(
```
tribeiro, line 637: I'd suggest adding a logger to the observatory model and using `self.log.warning` instead.. 

You could add something like this to the class, init method:

```
import logging

...
class Kinem_model(object):
    ...
    def __init__(
        self, location=None, park_alt=86.5, park_az=0.0, start_filter="r", mjd0=0, log=None
    ):

    self.log = log.getChild(type(self).__name__) if log is not None else logging.getLogger(type(self).__name__)
```

yoachim: Looks like a logger would then print the warning for every single exposure. I like the `warnings.warn` behavior of only printing once.

rubin_sim/scheduler/modelObservatory/kinem_model.py
```diff
@@ -104,6 +105,7 @@ def setup_camera(
         maxspeed=3.5,
         accel=1.0,
         decel=1.0,
+        two_motion_time=15.0,
```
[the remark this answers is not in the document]
yoachim: Let's try `shutter_2motion_min_time`

PR #182, opened 2022-04-11
U/lynnej/run galactic plane metrics

rhiannonlynne:
Modifies the minimum time interval for solar system metrics (per Mario's request). 
THIS IS IMPORTANT -- the previous solar system metrics on disk and in the summary files, do not use this same minimum time and so will not be directly comparable. It's possible that I should not merge this until we re-run the metrics with this update ..). 

Mostly, it adds a new metric to the metadata for the timespan in a single night, and makes sure to *actually* run the galactic plane metrics in the science bundle.

### DIFF
--- a/rubin_sim/maf/metrics/tgaps.py
+++ b/rubin_sim/maf/metrics/tgaps.py
@@ -7,6 +7,7 @@ __all__ = [
     "NightgapsMetric",
     "NVisitsPerNightMetric",
     "MaxGapMetric",
+    "NightTimespanMetric",
 ]
 
 
@@ -44,7 +45,7 @@ class TgapsMetric(BaseMetric):
         allGaps=False,
         bins=np.arange(0, 120.0, 5.0) / 60.0 / 24.0,
         units="days",
-        **kwargs
+        **kwargs,
     ):
         # Pass the same bins to the plotter.
         self.bins = bins
@@ -107,7 +108,7 @@ class TgapsPercentMetric(BaseMetric):
         minTime=2.0 / 24,
         maxTime=14.0 / 24,
         units="percent",
-        **kwargs
+        **kwargs,
     ):
         self.timesCol = timesCol
         assert minTime <= maxTime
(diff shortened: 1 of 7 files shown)


rubin_sim/maf/metrics/tgaps.py
```diff
@@ -255,3 +256,48 @@ def run(self, dataSlice, slicePoint=None):
...
+    def run(self, dataSlice, slicePoint=None):
+        data = np.sort(dataSlice, order=self.mjdCol)
+        unights, counts = np.unique(data[self.nightCol], return_counts=True)
+        unights = unights[np.where(counts > 1)]
+        if len(unights) == 0:
+            result = self.badval
+        else:
+            nstart = np.searchsorted(data[self.nightCol], unights, side="left")
+            nend = np.searchsorted(data[self.nightCol], unights, side="right") - 1
+            tspans = (data[self.mjdCol][nend] - data[self.mjdCol][nstart]) * 24.0 * 60.0
```
yoachim, line 301: Do you want to include cases where tspans=0? I'm not sure if it would be more informative clipping off instances where there wasn't a second observation or not.

yoachim (approved):
Looks good to me.

rubin_sim/maf/metrics/tgaps.py
rhiannonlynne: This is a good question. I thought briefly about it when creating the metric and decided that it was better to look at only cases where there are at least two visits .. it's not a super obvious answer though. You could definitely say that actually looking at any night where there was any visit was reasonable, and then using the total of all nights with visits as the basis for generating the X percentile to return. I don't think there is a clear answer, so for now, I'll leave it like this.

yoachim: I think the way the metric is written it is including nights with only one visit. The `unights` variable gets clipped, but `data` doesn't get the low count nights removed, so `tspans` is still going to have lots of zeros in it.

yoachim: No wait, it's doing it the right way. I always forget the order on `searchsorted`.

rhiannonlynne: But also - the `unights` are the nights where the count is >1, not just >0.

PR #184, opened 2022-04-26
Update archive for newer versions of json

rhiannonlynne:
Updates remote metrics, runs, and summaries file pointers
Tightens up plot limits in plot_run_metrics
Stops adding the baseline to the summary stats being plotted, if it's already in the list of runs.
Keeps Nans in plotted values, to stop confusing matplotlib.


ehneilsen (approved):
This all looks fine to me.

ehneilsen (approved):
looks  fine

PR #188, opened 2022-05-05
U/lynnej/cleanup archives

rhiannonlynne:
Minor updates to remove references to short_name_norm (which wasn't being used anyway), 
add some order tweaks to the microlensing metrics in the scienceRadar, 
tighten up plot limits on plot_run_metric if they were vertical (instead of just horizontal) for the 'run' values, 
and add a helper function to write an updated metric_sets.json file to disk.

### DIFF
--- a/bin/maf/gather_summaries
+++ b/bin/maf/gather_summaries
@@ -11,7 +11,7 @@ if __name__ == "__main__":
 
     parser = argparse.ArgumentParser(
         description="Find resultsDbs in a series of directories and "
-        "gather up their summary stats into a single CSV file. "
+        "gather up their summary stats into a single CSV or hdf5 file. "
         "Intended to run on a set of metrics run on multiple "
         "simulations, so that each resultsDb has similar summary"
         "statistics."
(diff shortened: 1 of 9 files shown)


bin/maf/gather_summaries
```diff
@@ -35,23 +35,41 @@ if __name__ == "__main__":
...
     mdict = rc.buildMetricDict()
-    print("len mdict", len(mdict))
+    print(f"And found {len(mdict)} metrics.")
     rc.addSummaryStats(mdict)
 
     # Save summary statistics
-    if args.suffix is None:
-        outfile = "summary.csv"
+    if args.to_pickle:
+        rc.summaryStats.to_pickle(outfile)
```
ehneilsen, line 72: I think maybe we should stay away from using pickles. They aren't really portable, not even between different versions of python. One (but not the only) better alternative is HDF5, which is both more portable and also much faster to write and read (even from python).

ehneilsen (approved):
looks fine.

bin/maf/gather_summaries
rhiannonlynne: Let me check with @yoachim - I added the pickle option because that's what he was using somewhere else and I didn't want to remove it if it was needed.

PR #189, opened 2022-05-10
U/yoachim/ddf batch fix

yoachim:
update to the new standard utility `ddf_locations`

### DIFF
--- a/rubin_sim/maf/batches/ddfBatch.py
+++ b/rubin_sim/maf/batches/ddfBatch.py
@@ -7,6 +7,20 @@ __all__ = ["ddfBatch"]
 
 
 def ddfBatch(runName="opsim", nside=512, radius=2.5, nside_sne=128):
+    """
+    Parameters
+    ----------
+    nside : int (512)
+        The HEALpix nside to run most of the metrics on. default 512.
+    radius : float (2.5)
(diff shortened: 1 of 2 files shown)


rhiannonlynne (approved):

rhiannonlynne:
Looks like there's an update to the GitHub actions (how it's installing conda, I think) required, but the changes shouldn't break anything anyway. 

It'd be nice if the resulting DDF metric names could be a little more standardized. Sometimes the filter is repeated in the output summary stat (and presumably the filenames created for the plots, etc.) and the handling of 'all' is a different pattern than the per-filters as well. 
If you used the info_label (which will override what is otherwise the default usage of the sql constraint for labelling different metric outputs), this could be more consistent. If you also used the `maf.batches.common.`filterList()`, you could get back a (zippable if needed) list of filters, and dictionaries (keyed by filter) of sql constraints (which can include the DDF selection criteria), standard plot colors per filter (for histograms), and corresponding easier to read info_labels, which would let the per-filter results come up on the showMaf pages in filter order bluer to redder. It's not really a big deal, but does mean that I'm doing a fair bit of remapping in the summary metrics comparisons to more readable names at that point.

rhiannonlynne:
Oh - sorry about 'closing' the PR. I think I must have hit a button by accident.

PR #190, opened 2022-05-11
Update YoungStellarObjectsMetric.py

lprisinzano:
I found a small error at line 82. i=7.97 is the correct value. The value 7.37 was referred to mag z that is not used in this metrics

### DIFF
--- a/rubin_sim/maf/mafContrib/YoungStellarObjectsMetric.py
+++ b/rubin_sim/maf/mafContrib/YoungStellarObjectsMetric.py
@@ -79,7 +79,7 @@ class NYoungStarsMetric(BaseMetric):
         m5Col="fiveSigmaDepth",
         filterCol="filter",
         badval=0,
-        mags={"g": 10.32, "r": 9.28, "i": 7.37},
+        mags={"g": 10.32, "r": 9.28, "i": 7.97},
         galb_limit=90.0,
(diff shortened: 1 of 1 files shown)


rhiannonlynne:
Thanks Loredana! 
I have a quick question: how much of an impact on the results of the metric does this update make? 

We would prefer not to rerun all of the science metrics on all of the runs at this time, so if it's a minor change in the result (or a change that doesn't significantly impact the relative ranking of different survey strategies, that would be very helpful to know.

rhiannonlynne:
I'm going to merge this onto a branch where I'm staging some metric updates, so we can contain all of the changes to the scienceRadar metrics in one spot, so it's easier to control when we include and rerun them.

lprisinzano:
Dear Lynne,
I just checked within DataLab and compared the results using both i=9.79 and i=9.37 
and the results are the same. On the contrary, I tried to change the g mag limit by using two different values 
and I found a difference. This means that changing i from 7.37 to 7.97 has not impact since the metrics
computes the maximum distance we can achieve by using simultaneously g, r and i,  and the computation is not sensitive to a change of about 0.5 mag 
while it is for a change in the g magnitude.  So no need to rerun all of the science metrics but I would prefer to
include in the code the right value i=7.97. 


Loredana

PR #192, opened 2022-05-17
U/lynnej/display dict fixes

rhiannonlynne:
Update the science radar batch with :
* display dict fixes where captions were incorrect
* extend the periodModulationMetric to 10 years
* fix the I band depth for the YoungStellarObjectsMetric


yoachim (approved):
That all looks good

rhiannonlynne:
Cool. I guess I'd like to make sure we're ready to use this before merging it -- although, there is also now a conda tag at what I believe was used to generate the previous scienceRadar outputs, so it is not critical I suppose.

rhiannonlynne:
Looks like the NGals scaling metric wasn't using a slicer with the cache turned off - I updated that, so we may see some changes in those results. (expect it's minor, given the dust cutoff).

rhiannonlynne:
Ah. It looks like I failed to add the LVDwarfs metric to the science radar batch.
Adding that as well.

PR #194, opened 2022-05-25
U/yoachim/sky2hdf5

yoachim:
* Update the pre-computed skybrightness files to use hdf5 rather than npz. This let's one quickly load a small timescale from disk rather than the entire file. 
* Removed pre-computing things like airmass, sun moon planet positions, and zenith mask. Results in much smaller save files.


rhiannonlynne:
It looks like ephem could now be dropped from the requirements.txt file and that the GitHub actions need to be updated as the skybrightness_pre_small data is now irrelevant. 
We should also tag the repo after committing this, and update the conda build to remove ephem from those build requirements (the conda feedstock tracks requirements separately).

rhiannonlynne (approved):

PR #196, opened 2022-06-09
Create filterPairTGapsMetric.py

xiaolng:
metric to evaluate the coverage of time gaps in filter pairs;

### DIFF
--- a/rubin_sim/maf/mafContrib/filterPairTGapsMetric.py
+++ b/rubin_sim/maf/mafContrib/filterPairTGapsMetric.py
@@ -1,0 +1,175 @@
+import numpy as np
+#from lsst.sims.maf.utils import radec2pix
+
+# if rubin_sim installed
+from rubin_sim.maf.utils.mafUtils import radec2pix
+import rubin_sim.maf.metrics as metrics
+
+class filterPairTGapsMetric(metrics.BaseMetric):
+    """
+    Parameters:
+        colname: list, ['observationStartMJD', 'filter', 'fiveSigmaDepth']
+        fltpairs: filter pair, default ['uu', 'ug', 'ur', 'ui', 'uz','uy',
+                                        'gg', 'gr', 'gi', 'gz', 'gy',
+                                        'rr', 'ri', 'rz', 'ry', 
+                                        'ii', 'iz', 'iy', 
+                                        'zz', 'zy',
+                                        'yy']
+        mag_lim: list, fiveSigmaDepth threshold each filter, default {'u':18, 'g':18, 'r':18, 'i':18, 'z':18, 'y':18}
+        bins_same: np.array, bins to get histogram for same-filter pair ;
+        bins_diff: np.array, bins to get histogram for diff-filter pair ;
+        save_dT: boolean, save time gaps array as result if True
+        allgaps: boolean, all possible pairs if True, else consider only nearest 
+        
+    Returns:
+        result: dictionary, 
+        reduce_tgaps: median
+        
+    """
+
+    def __init__(self, colname=['observationStartMJD', 'filter', 'fiveSigmaDepth'], 
+                 fltpairs= ['uu', 'ug', 'ur', 'ui', 'uz','uy', 'gg', 'gr', 'gi', 'gz', 'gy', \
+                             'rr', 'ri', 'rz', 'ry', 'ii', 'iz', 'iy',  'zz', 'zy', 'yy'],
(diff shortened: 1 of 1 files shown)


rubin_sim/maf/mafContrib/filterPairTGapsMetric.py
```diff
@@ -0,0 +1,175 @@
...
+        return dT
+    
+    def run(self, dataSlice, slicePoint=None):
+                        
+        # sort the dataSlice in order of time.  
+        dataSlice.sort(order='observationStartMJD')
+        
+        fieldRA = np.mean(dataSlice['fieldRA']) 
+        fieldDec = np.mean(dataSlice['fieldDec'])
+        pixId = radec2pix(nside=16, ra=np.radians(fieldRA), dec=np.radians(fieldDec))
```
rhiannonlynne, line 130: Hi @xiaolng - I think you actually mean to get these from the slicePoint, not the DataSlice. 
The dataSlice will have a variety of values for fieldRA and fieldDec (because we dither). However, what I think you're trying to evaluate is what are the visits which are relevant for *this point on the sky* -- aka, the slicePoint values. 
If you replace 
```fieldRA = np.mean(dataSlice['fieldRA'])```
with 
```ra = slicePoint['ra'] ```
you will get the point on the sky that you actually mean to evaluate (and avoid any problems that may come from wrap-around at RA=0, where the mean value of a field which has a central pointing at say RA=0.2 and another field with a central pointing of say RA=359.2 would be 180 or so .. and a totally different part of the sky. 

Then again -- you don't even need to calculate epixID, if you are looking for the healpix ID of this point on the sky. You can use slicePoint['sid'] for the 'sliceId'. 

Let me suggest though, that you don't even need that for this metric. The values that are returned by the metric are saved in an array that will match the healpix ids .. so just looking at what index value in the metricValues is being used == this is the healpix id.

File: rubin_sim/maf/mafContrib/filterPairTGapsMetric.py [hunk 2]
```diff
@@ -0,0 +1,175 @@
...
+                'dT_dic': dT_dic,
+                'fom_dic': fom_dic,
+                'dataSlice': dataSlice
+                  }
+        else:
+            result = {
+                'pixId': pixId,
+                'fom_dic': fom_dic,
+                  }
+        return result
```
rhiannonlynne, line 170: Okay, so I thought when we talked, we had discussed this metric actually providing a single value per healpix point? It looks like you are gearing up here to return the clipped value per filter pair. 
Are you intending to run this yourself and save the full outputs (and then investigate them)? If so, this makes sense to leave in place (although I would suggest you NEVER want to return the dataSlice, as you can recreate it easily enough from the slicer and it's a TON of data .. basically replicating every visit for every point in the healpix slicer). 
I would wonder if perhaps the extra memory overhead slows this down? 

I would gently suggest: 
If you're feeling like you've debugged this metric, drop the first "if self.save_dT" option. (if you need to debug in the future, you can always modify the metric in place and just return all of these values, then go back to "normal"). 
You do not need to return the pixId, it is preserve in other places in the metricValues and slicer combination.
Possibly consider only returning the contents of the 'reduce_tgaps_FoM' in place of the fom_dic, if you are not planning on running this on your own (and saving the output) .. we won't be saving the full outputs and you will not have this information available, so it just ends up increasing the memory footprint of the metric unnecessarily.

rhiannonlynne, line 170: Okay, so I thought when we talked, we had discussed this metric actually providing a single value per healpix point? It looks like you are gearing up here to return the clipped rms histogram value per filter pair. 

I would gently suggest: 
If you're feeling like you've debugged this metric, drop the first "if self.save_dT" option. (if you need to debug in the future, you can always modify the metric in place and just return all of these values, then go back to "normal". You probably do not need to return the dataSlice values, as you can recreate those from the slicer). 

You do not need to return the pixId, it is preserve in other places in the metricValues and slicer combination.

Possibly consider only returning the contents of the 'reduce_tgaps_FoM' in place of the fom_dic, if you are not planning on running this on your own (and saving the output) .. we won't be saving the full outputs and you will not have this information available, so it just ends up increasing the memory footprint of the metric unnecessarily.

rhiannonlynne (changes requested):
At least the pixId calculation should be updated, preferably removed entirely as unnecessary, the removal of the other pieces are updates I would like to see but you may have needs for them I don't understand.

rhiannonlynne (approved):
Thank you!

rhiannonlynne (approved):

PR #198, opened 2022-06-17
Tickets/preops 1256

ehneilsen:
Utilities to make it easier to write plotting software, and interact with in jupyter notebooks.


yoachim (approved):
That all seems in order.

PR #200, opened 2022-06-21
update data file and add test

yoachim:
Fix broken symlink. Add super basic unit tests to catch it in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ To update the source contents of the data files:
 
 * Update the files in your local installation
 * If you are updating the baseline sim, create a symlink of the new database to baseline.db
-* Create a new tar file with a new name, e.g., `tar -chvzf maf_june_2021.tgz maf`
+* Create a new tar file with a new name, e.g., `tar -chvzf maf_2021_06_01.tgz maf` (no `-h` if symlinks should stay as symlinks)
 * Copy your new tar file to NCSA lsst-login01.ncsa.illinois.edu:/lsstdata/user/staff/web_data/sim-data/rubin_sim_data/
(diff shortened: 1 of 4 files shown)


rhiannonlynne:
Can I ask what was broken? The tar file at NCSA or something else? 
I found that on Datalab, the sim_baseline data directory ended up containing a baseline.db symlink that turned into not-a-symlink .. but it was because the data directory was copied (outside of rubin_sim downloading it). 
I think a test to catch if that has happened is good, but won't solve the problem of sometimes people copy the sim_baseline directory and then it's a broken symlink. (because when rubin_sim downloaded the data, it was fine).

yoachim:
Are you sure it was a copy that turned the symlink into a file? Because right now the current tar file has a file instead of a symlink so `get_baseline` won't work with a fresh install. I guess I could change it to just use `glob` and then there's no symlink issue at all.

rhiannonlynne:
Ah - I see why the unit test for the test_data is failing. 
Question: what is the purpose of the unit test? Are these run on a conda install? I am not sure. 
They *would* be run in GitHub when updating the baseline data directory .. but these basic unit tests currently don't download the sim_baseline data set, because it's not needed for any other unit tests. 
Does sim_baseline need to be added to the downloaded test data, for the shorter unit tests we're running here? It is downloaded and available for the full tests .. we could, say, require the full tests to be run when tagging the repo? (I'd have to figure out the GitHub action to make that work, but it should be doable). Or is it enough to trust that the glob takes care of the issue we were having, and set this test to only run if the sim_baseline directory is present?

PR #201, opened 2022-06-22
Tickets/dm 35108

tribeiro:
* Replace `setup.py` with `pyproject.toml`. 
   A coupe things to notice:
   * We still retain a minimum `setup.py`. This is still needed to get the "dynamic version" for the package as pyproject.toml alone does not support it. It seems like support is coming in the future which would allow us to remove `setup.py` completely.
   * Most of the content from `setup.cfg` got moved to `pyproject.toml`. Nevertheless, flake8 does not support `pyproject.toml` and the developer is not interested in adding support. So we still need a minimum `setup.cfg` to configure flake8.
   * Executables are now handled by entry points, defined in the `pyproject.toml` file. When we install the package with `pip` or conda, the executables are generated automatically.  
*  Add `BalanceVisits` basis functions. This is a basis function is used to group together different surveys and balance the observations such that a particular survey is penalized if it has more observations than the others in the group. This is used mostly to work around how we do imaging survey with the Auxiliary Telescope now.
* Add `RewardNObsSequence` basis function. This basis function rewards staying in a survey when it is composed by more than one observation. It is also used on AT to take sequence of observations in different filters of the same survey/field.


rubin_sim/scheduler/basis_functions/basis_functions.py
```diff
@@ -2004,3 +2005,34 @@ def _calc_value(self, conditions, indx=None):
...
+    def __init__(self, n_obs_survey, note_survey, nside=None):
+        super().__init__(nside=nside)
+
+        self.n_obs_survey = n_obs_survey
+
+        self.survey_features = {}
+        self.survey_features["N_obs_survey"] = features.N_obs_survey(note=note_survey)
+
+    def _calc_value(self, conditions, indx=None):
+        return self.survey_features["N_obs_survey"].feature % self.n_obs_survey
```
rhiannonlynne, line 2038: How does this reward the observations in a survey being taken together? Maybe I am just not familiar with the current reward functions, etc. n_obs_survey = 'number of observations to reward' .. I would have thought this meant that the observations were sequential or maybe that once you had this many observations in this survey, that that number of observations was what you wanted to reward? but I am not sure?)
Maybe what I'm looking for here is a way to understand how to combine this with other reward functions, such as slew time, etc, if you're coming into the scheduler and seeing this function. 
(note: I think this kind of information is probably lacking all over the scheduler, but is something that seems like it would be helpful to start including).

rubin_sim/scheduler/basis_functions/basis_functions.py
```diff
@@ -1949,3 +1950,57 @@ def _calc_value(self, conditions, indx=None):
...
+            1
+            + np.floor(
+                self.survey_features["N_obs_survey_interest"].feature
+                / self.nobs_reference
+            )
+        ) / (
+            self.survey_features["N_obs_survey"].feature
+            if self.survey_features["N_obs_survey"].feature > 0
+            else 1
+        )
```
rhiannonlynne, line 2007: Somehow, this sounds very similar to the other basis function added - RewardNObsSequence - so I guess I don't understand how it's different or the same?

pyproject.toml
```diff
@@ -0,0 +1,69 @@
...
+license = { text = "GPL" }
+classifiers = [ 
+    "Intended Audience :: Science/Research",
+    "License :: OSI Approved ::  GNU General Public License v3 or later (GPLv3+)",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.7",
+    "Topic :: Scientific/Engineering :: Astronomy",
+ ]
+urls = {documentation = "https://jira.lsstcorp.org/secure/Dashboard.jspa", repository = "https://github.com/lsst/rubin_sim" }
```
rhiannonlynne, line 17: Should the documentation link be rubin-sim.lsst.io?

pyproject.toml
```diff
@@ -0,0 +1,69 @@
...
+rs_download_data = "rubin_sim.data.rs_download_data:rs_download_data"
+rs_download_sky = "rubin_sim.data.rs_download_sky:rs_download_sky"
+addRun = "rubin_sim.maf.db.add_run:addRun"
+ddf_dir = "rubin_sim.maf.ddf_dir:ddf_dir"
+gather_summaries = "rubin_sim.maf.runComparison.gather_summaries:gather_summaries"
+generate_ss = "rubin_sim.maf.generate_ss:generate_ss"
+glance_dir = "rubin_sim.maf.glance_dir:glance_dir"
+maf_night_report = "rubin_sim.maf.maf_night_report:maf_night_report"
+metadata_dir = "rubin_sim.maf.metadata_dir:metadata_dir"
+resultdb_combine = "rubin_sim.maf.runComparison.resultdb_combine:resultdb_combine"
```
rhiannonlynne, line 30: I think we can drop this one - I thought it had already been removed from the bin directory. It does the same thing as gather_summaries, but without as many options.

pyproject.toml
```diff
@@ -0,0 +1,69 @@
...
+run_moving_calc = "rubin_sim.maf.run_moving_calc:run_moving_calc"
+run_moving_fractions = "rubin_sim.maf.run_moving_fractions:run_moving_fractions"
+run_moving_join = "rubin_sim.maf.batches.run_moving_join:run_moving_join"
+scimaf_dir = "rubin_sim.maf.scimaf_dir:scimaf_dir"
+showMaf = "rubin_sim.maf.show_maf:showMaf"
+addSeds = "rubin_sim.movingObjects.add_seds:add_seds"
+generateCoefficients = "rubin_sim.movingObjects.generate_coefficients:generate_coefficients"
```
rhiannonlynne, line 37: We. can drop this one too.

rhiannonlynne: (as in we can drop addSeds and also drop generateCoefficients).

rhiannonlynne (approved):
It would be lovely to figure out how we need to document basis functions so people in general can use them without writing them and experimenting with them from scratch. 

The other parts of the PR that deal with moving to pyproject.toml -- looks good. I realized after commenting that some of the scripts could be dropped that maybe what I was thinking is that they could be dropped from the entry points / bin script .. but, I suspect having the ability to use the code within each of those as a function is likely useful (for generateCoefficients and addSeds -- the resultsdb_combine is redundant and we should just drop it entirely). 

Generally looks good. I am curious if the requirements need to go into the pyporject.toml file now? (they seem to still be floating in a requirements.txt file and they still exist separately in our conga-forge feedstock repo as well -- does pyproject.toml combine these?)

rhiannonlynne:
Hi Tiago - I think the branches that did some work on other bin scripts are merged in, so you could go ahead and rebase/be done. The other pending branches/PRs don't touch those bin scripts, I believe.

rubin_sim/maf/batches/run_moving_join.py
```diff
@@ -8,12 +8,11 @@
```
rhiannonlynne, line 8: Curious - why did this one (run_moving_join.py) end up in rubin_sim/maf/batches, while the other similar scripts (run_moving_calc.py, for example) end up in rubin_sim/maf ?

rhiannonlynne (approved):

rubin_sim/maf/glance_dir.py
```diff
@@ -12,7 +12,7 @@
 import argparse
 
 
-if __name__ == "__main__":
+def glance_dir():
     """
```
ehneilsen, line 16: I think it is usually preferred style not to put a newline right after the opening triple-quote, but to put the summary on the same line as it.

ehneilsen (approved):
If the pyproject.toml file is supposed to handle dependencies, perhaps the dependencies in requirements.txt should be added to it?

rubin_sim/scheduler/basis_functions/basis_functions.py
tribeiro: So, I know they sound similar but they are actually pretty complementary. This basis function is for the cases when a single survey observation is comprised of multiple observations. Think Deep Drilling surveys. On DD this is not really an issue because their rewards is either -inf or +inf (or something like that). But the survey I use for AuxTel is a small variation of DD where the reward is actually a value. And we merge observations in different filters, same way we normally do with DD. But, because this survey is a value, it happens sometime that the value of another survey wins so we don't complete the observations. This basis function gives a boost in the reward once you did 1 observation and continues until to completes the total number of observations.

The other basis function os more on tying together observations of different surveys.

pyproject.toml
```diff
@@ -0,0 +1,69 @@
...
+write_to_template = """
+# Generated by setuptools_scm
+__all__ = ["__version__"]
+__version__ = "{version}"
+"""
+
+[tool.pytest.ini_options]
+addopts = "--black --ignore-glob=*/version.py"
+flake8-ignore = ["E133", "E203", "E226", "E228", "N802", "N803", "N806", "N812", "N813", "N815", "N816", "W503"]
+flake8-max-line-length = 110
```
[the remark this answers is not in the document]
tribeiro: so, black does not break down strings (rightly so). So you can still, in principle, end up with lines that are longer than what black tries to wrap them up, and that is also fine, as long as they are less than `flake8-max-line-length`.

rubin_sim/maf/batches/run_moving_join.py
tribeiro: I think I was trying to place the method were they seem like they belonged.. but I am not sure I got all of them in the correct place..

tribeiro:
> If the pyproject.toml file is supposed to handle dependencies, perhaps the dependencies in requirements.txt should be added to it?

@ehneilsen so, the conda dependencies are not read from requirements.txt or the new `pyproject.toml`. They are actually on the [conda recipe](https://github.com/conda-forge/rubin-sim-feedstock/blob/main/recipe/meta.yaml). 

Now we still use the `requirement.txt` file for people to setup the package for local development, so I kept it here. I am not sure we could move that to `pyproject.toml`, especially since we suggest using `conda` to install those dependencies.

PR #197, opened 2022-06-17
code checked and updated

pgris:
Few modifications/checks following the alleged discrepancy of this metric.

PR #203, opened 2022-06-23
NSN code check and update

rhiannonlynne:
This is the continuation of the PR from @pgris (https://github.com/lsst/rubin_sim/pull/197)
We merged that PR to a branch, to address some unresolved questions, be able to reformat the files (black), and to address some run-time issues @yoachim was investigating.


rhiannonlynne:
I am going to ping Philipe Gris @pgris  and Federica Bianco (@fedhere) in this PR. 
Philipe - if you have additional changes to make to the code, could you please target this branch in any additional PRs? And also if you do a 'git pull' in your fork and change to this branch prior to that, it will make the process easier.

rhiannonlynne:
@pgris did you look into calculating magnitudes, etc. for the SN in a multi-dimensional array instead of per-band? It seems like this is the only use of the multiprocessing in this metric, and so I was wondering if you'd looked at calculating these in multi-D arrays within numpy instead. 

Also - pandas dataframes do tend to be a bit slower, especially when creating many of them. I see a few things that are more easily available in pandas (primarily group by + apply), but have not looked at speed differences with numpy + where. Did you already look into these differences as well?

yoachim (approved):
OK, I've managed to remove a net of about 1k lines of code. There' still a little behavior in the unit test I'm confused about, and I suspect we can further optimize some of the pandas operations, but this is a good start for now.

I'll update the data files at NCSA then merge.

PR #205, opened 2022-06-29
Make str representation work even when the queue is None

ehneilsen:
The new __str__ was failing when the queue was None. Fix that.


ehneilsen:
Somehow I messed up this request and included a bunch of changes that weren't supposed to be there. Closing the request, will try again.

PR #206, opened 2022-06-30
U/yoachim/mjd0

yoachim:
Scrub out all the places we had hard coded the survey start date and replace with a single utility so everything gets updated in one place.

### DIFF
--- a/rubin_sim/maf/mafContrib/TDEsPopMetric.py
+++ b/rubin_sim/maf/mafContrib/TDEsPopMetric.py
@@ -85,10 +85,7 @@ class TdePopMetric(metrics.BaseMetric):
         self.ptsNeeded = ptsNeeded
 
         self.lightcurves = Tde_lc(file_list=file_list)
-        if mjd0 is None:
-            self.mjd0 = survey_start_mjd()
-        else:
-            self.mjd0 = mjd0
+        self.mjd0 = survey_start_mjd() if mjd0 is None else mjd0
 
         dust_properties = Dust_values()
         self.Ax1 = dust_properties.Ax1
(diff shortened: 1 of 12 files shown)


rubin_sim/maf/mafContrib/TDEsPopMetric.py
```diff
@@ -86,7 +85,10 @@ def __init__(
         self.ptsNeeded = ptsNeeded
 
         self.lightcurves = Tde_lc(file_list=file_list)
-        self.mjd0 = mjd0
+        if mjd0 is None:
+            self.mjd0 = survey_start_mjd()
+        else:
+            self.mjd0 = mjd0
```
ehneilsen, line 91: This is a classic case where a ternary operator might be good:
`self.mjd0 = survey_start_mjd() if mjd0 is None else mjd0`

Really, though, it doesn't matter.

rubin_sim/maf/mafContrib/TDEsPopMetric.py
```diff
@@ -225,7 +227,6 @@ def __init__(
         )
 
     def _some_color_pnum_detect(self, dataSlice, slicePoint, mags, t):
-        result = 1
```
ehneilsen, line 228: Don't know what dropping this line does, of if it has anything to do with the start mjd.

ehneilsen (approved):
I don't understand why you get the start MJD from a function rather than just importing the module variable that the function returns, which just seems to add an unnecessary extra level of indirection. Do we think we might want to have the code calculate the start date "on the fly" in the future?

Also, there were a couple of changes that seemed unrelated to the start mjd, which I didn't understand. (See the comments on the lines themselves.)

rubin_sim/maf/mafContrib/TDEsPopMetric.py
yoachim: It was defined and never used, so I just sniped it on the way by.

rubin_sim/maf/mafContrib/TDEsPopMetric.py
yoachim: I like it!

PR #209, opened 2022-07-05
Make env a global variable

rhiannonlynne:
Showmaf wasn't working .. now it is.
This  fix however, does feel like a bit of a hack. 
Making a bunch of variables 'global' (note that it's not just `env` .. I just added `env` to the set of things which were defined as global variables) seems like not the optimal approach, although it works.


yoachim (approved):
seems like the simple fix is the best fix to me.

PR #210, opened 2022-07-10
Debug XRB metric

ebellm:
@rhiannonlynne pointed out that we were randomly generating unphysical negative orbital periods, which created warnings.  This PR fixes that problem and improves the docstrings.


rhiannonlynne:
Thanks Eric! Lovely docstrings, btw.

PR #212, opened 2022-07-19
U/lynnej/update season calc

rhiannonlynne:
I went through and tested this more carefully, looked at many examples. 
Current "season" calculations based on gaps in the nights between visits are likely to fail, based on this examination (there can be gaps of 150 nights or so in infrequently visited portions of the sky). The new, more precise numbers for the season calculation no longer assign visits into the wrong season (this sometimes happened at the end of seasons, and likely at preferential RA values, as essentially the time of the sync-up of the seasons was slightly wrong). 
Notebook demonstration at https://github.com/lsst/rubin_sim_notebooks/blob/main/maf/science/Test%20Seasons.ipynb


yoachim (approved):
Looks like this is a pretty minor change.

PR #214, opened 2022-07-20
U/pgris/nsnmetric

rhiannonlynne:
Updates to SN metric.

I kind of think we'll have some more PRs to this branch, but I'd like to make sure we get this in a place where it may be easier to view our updates for @pgris .


rhiannonlynne:
My estimate for the time required to run this metric at nside=16 is between 6 to 7 hours.

pgris:
This is strange. I ran the notebook SNIa.ipynb on baseline_v2.1_10yrs and it took about 3 hours with nside=16 and the parameters zStep=0.03, daymaxStep=0.03,  dust=False, hard_dust_cut=0.25, n_bef=3, n_aft=8.

rhiannonlynne:
I see. This was also my rough estimate based on extrapolating the timing from a subset of points. When I ran on the full sky at nside=16, with those parameters, I found the time required to be 3h 51min 4s, on my laptop. I'll see how it does on our cluster. 
(daymaxStep = 3, not 0.03 yes?)

PR #215, opened 2022-07-27
U/yoachim/rad deg fix

yoachim:
Fixed a random minor degrees/radians bug. Put the `norm_df` utility back it since I was using it and it got deleted. Fixed a bug generating captions.


ehneilsen (approved):
Looks fine, as far as I can tell.

PR #216, opened 2022-07-28
U/yoachim/readme update

yoachim:
update README. Add links to where folks can get help. Remove goto statements in install instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ make html
 
 ## Getting Help ##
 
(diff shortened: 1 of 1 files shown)


rhiannonlynne (approved):
Looks good! Thank you. 
I worry that we may miss messages that are just generally posted in community.lsst.org, yet I also don't want to say they must be posted in survey strategy or necessarily the 'science' categories. Maybe we put our tags on community here and ask them to tag at least one of us? (wonder if community could add a tag that would notify us without being specific to our names).

PR #218, opened 2022-08-01
U/lynnej/cleanup misc

rhiannonlynne:
Biggest functional difference is to allow both invert+mag in normalize_metric_summaries.


ehneilsen (approved):
Looks fine.

PR #219, opened 2022-08-02
U/pgris/nsnmetric

pgris:
New version of the code, with telescope parameters embedded in the gamma*.hdf5 file to avoid uncontrol use of these parameters by the user. Using the notebook SNIA.ipynb lead to the following results, with the metric parameters (zStep=0.03,daymaxStep=3,n_bef=3,n_aft=8,zmin=0.1,zmax=0.5,snr_min=1.): and nside=16:
         OS                                              |  processing time (s) |    NSN     | zcomplete |
baseline_v2.1_10yrs                           |               8681           |  92642    |    0.276     |
rolling_bulge_ns2_rw0.9_v2.0_10yrs |               8868           |   91578   |    0.272      |
baseline_v2.0_10yrs                           |               8829           |  92217    |       0.274    |

### DIFF
--- a/rubin_sim/maf/metrics/snNSNMetric.py
+++ b/rubin_sim/maf/metrics/snNSNMetric.py
@@ -36,7 +36,11 @@ class SNNSNMetric(BaseMetric):
     nexpCol : str,opt
       number of exposure column name (default : numExposures)
      vistimeCol : str,opt
-        visit time column name (default : visitTime)
+        visit time column name (default : visitTime)   
+    seeingCol : str,opt
+         seeing column name (default: seeingFwhmEff)
+    noteCol: str, opt
+         note column name (default: note)
     season : list,opt
        list of seasons to process (float)(default: -1 = all seasons)
     coadd : bool,opt
@@ -66,7 +70,13 @@ class SNNSNMetric(BaseMetric):
     dust : bool (True)
       Apply dust extinction to visit depth values
     hard_dust_cut : float (None)
-      If set, cut any point on the sky that has an ebv extinction higher than the hard_dust_cut value.
+      If set, cut any point on the sky that has an ebv extinction higher than the hard_dust_cut value.  
+    telescope_params: dict,opt
+      telescope parameters (zp and mean wave lenghts per band)
+    DD_list: list(str), opt
+      list of DD runs in simu db
+    fieldType: str, opt
+      type of field to process (WFD or DD)(default: WFD)
     """
 
     def __init__(
@@ -81,17 +91,18 @@ class SNNSNMetric(BaseMetric):
         nexpCol="numExposures",
         vistimeCol="visitTime",
         seeingCol="seeingFwhmEff",
+        noteCol="note",
         season=[-1],
         coadd_night=True,
-        zmin=0.0,
-        zmax=1.2,
+        zmin=0.1,
+        zmax=0.5,
         zStep=0.03,
         daymaxStep=3.0,
         verbose=False,
         ploteffi=False,
         n_bef=3,
         n_aft=8,
-        snr_min=5.0,
+        snr_min=1.0,
         n_phase_min=1,
         n_phase_max=1,
         sigmaC=0.04,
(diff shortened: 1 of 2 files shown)


rubin_sim/maf/metrics/snNSNMetric.py
```diff
@@ -100,6 +107,9 @@ def __init__(
         add_dust=False,
         hard_dust_cut=0.25,
         gammaName="gamma_WFD.hdf5",
+        DD_list=['DD:COSMOS', 'DD:ECDFS', 'DD:EDFS, a', 'DD:EDFS, b', 'DD:ELAISS1',
+                 'DD:XMM-LSS'],
+            fieldType='WFD',
```
rhiannonlynne, line 112: Why are these specified here? What is the goal?

rubin_sim/maf/metrics/snNSNMetric.py
```diff
@@ -236,6 +234,13 @@ def run(self, dataSlice, slicePoint):
         metricVal : `np.recarray`
             ['nSN', 'zlim'] at this point on the sky
         """
+        # remove or keep DD runs depending on fieldType to process
+
+        DDobs = np.in1d(dataSlice[self.noteCol], self.DD_list)
+        if self.fieldType == 'DD':
+            dataSlice = dataSlice[DDobs]
+        else:
+            dataSlice = dataSlice[~DDobs]
```
rhiannonlynne, line 243: I understand why you have included this here. 
Typically we take care of this in the calling script (removing or allowing observations based on the "note" field is a sqlconstraint kind of thing for the MetricBundle).
If you need us to leave this here, so that your copy of the code can track ours, we can do that. 
I would note that the list of DD fields can potentially change, so you would need to modify the calling script and metric configuration in that case.

rubin_sim/maf/metrics/snNSNMetric.py
```diff
@@ -81,17 +87,18 @@ def __init__(
         nexpCol="numExposures",
         vistimeCol="visitTime",
         seeingCol="seeingFwhmEff",
+        noteCol="note",
         season=[-1],
         coadd_night=True,
-        zmin=0.0,
-        zmax=1.2,
+        zmin=0.1,
+        zmax=0.5,
```
rhiannonlynne, line 94: This should help speed up the metric calculation, if a maximum z limit of 0.5 is acceptable. I take it that this is the maximum value you've found when running the metric on simulations, thus the update? 
Does it need to be different for DD fields?

rubin_sim/maf/metrics/snNSNMetric.py
```diff
@@ -81,17 +87,18 @@ def __init__(
...
         zStep=0.03,
         daymaxStep=3.0,
         verbose=False,
         ploteffi=False,
         n_bef=3,
         n_aft=8,
-        snr_min=5.0,
+        snr_min=1.0,
         n_phase_min=1,
         n_phase_max=1,
```
rhiannonlynne, line 103: Can you make a comment about what snr_min means? Usually I would interpret this as saying that a minimum SNR of 1 is required for observations; that must be incorrect here though, so I am curious what this is intended to mean.

rhiannonlynne (commented):
Could you please provide some comments about the questions below?

rubin_sim/maf/metrics/snNSNMetric.py
pgris: The list od DD fields is an input parameter of the metric. So it can be changed when necessary. I have added these lines because the templates of gamma_WFD.hdf5 are not valid for all DDFs. So I thought that it would be better to be sure that the processing is done either for WFD or for DDFs.

rubin_sim/maf/metrics/snNSNMetric.py
pgris: Yes, 0.5 should be the max for WFD runs and this value has been used for the results quoted in the PR (processing time) . For DDFs, one should set: zmin=0.1, zmax=1.0

rubin_sim/maf/metrics/snNSNMetric.py
pgris: snr_min is the minimum SNR of the light curve points used perform selections (n_min_phase, n_max_phase, number of epochs before and after max) and to estimate sigmaC.

rhiannonlynne: Hmm. Is this compatible with the fact that DM is only expecting to measure and release alerts on point sources with SNR=5 or higher? It could be, if forced photometry measurements of the source's location on the sky, which I think should be acquired after the source reaches snr=5, are useful.

rubin_sim/maf/metrics/snNSNMetric.py
rhiannonlynne: Thank you for the update for DDF fields.

rubin_sim/maf/metrics/snNSNMetric.py
pgris: The metric assumes that the data are not coming from alerts but from image processing. We hope that SNR=1 will be possible.

PR #220, opened 2022-08-02
Make sure basis functions have unique labels

ehneilsen:


yoachim:
That seems fine. I'm not sure that guarantees basis functions will have unique names (maybe they will within a survey object). I imagine you would want to build a full label that has the basis function label and the survey object label. That would probably be unique.

ehneilsen:
Within a survey object is all that matters for my use, but I think my inclusion of id(basis_function)  (which gives the memory address of the instance of the basis function class) should be globally unique, if hacky and a bit opaque to the user. I initially wanted to do it using members of the various basis function subclasses, but discovered that sometimes a survey had multiple instances of the same basis function will all parameters identical, so this doesn’t work. (An example of this is the blob survey in which both filters are “y”.)


From: Peter Yoachim ***@***.***>
Date: Tuesday, August 2, 2022 at 6:07 PM
Subject: Re: [lsst/rubin_sim] Make sure basis functions have unique labels (PR #220)


That seems fine. I'm not sure that guarantees basis functions will have unique names (maybe they will within a survey object). I imagine you would want to build a full label that has the basis function label and the survey object label. That would probably be unique.

yoachim (approved):

PR #221, opened 2022-08-04
U/yoachim/satellites

yoachim:
Some tools for looking at the impact of satellite mega constellations. Not sure how to get the CI to pick up the new requirements, tried adding them to `test-requirements.txt` and `requirements.txt` but they don't get picked up.

### DIFF
--- a/tests/satellite_constellations/test_satellites.py
+++ b/tests/satellite_constellations/test_satellites.py
@@ -11,7 +11,7 @@ from rubin_sim.satellite_constellations import (
 
 class TestSatellites(unittest.TestCase):
     def test_constellations(self):
-        """Test instantiation of slicer sets slicer type as expected."""
+        """Test stellite constellations"""
 
         mjd0 = survey_start_mjd()
         sv1 = starlink_tles_v1()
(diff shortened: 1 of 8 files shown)


rhiannonlynne:
I'm not sure how the lsst-ts Jenkins requirements are set. It's possible it's using our requirements.txt file, but rather a proper conda "meta.yaml" recipe, so it would have to be updated there. (note that you will also have to update the meta.yaml in OUR condo-forge version of the feedstock).

tests/satellite_constellations/test_satellites.py
```diff
@@ -0,0 +1,35 @@
...
+    starlink_tles_v2,
+    oneweb_tles,
+    Constellation,
+)
+
+
+class TestSatellites(unittest.TestCase):
+    def test_constellations(self):
+        """Test instantiation of slicer sets slicer type as expected."""
+
```
rhiannonlynne, line 15: Please update the docstring.

rhiannonlynne:
I was going to ask why you would subclass the model_observatory instead of adding satellite information into the original satellite observatory (and leaving the utilities to feed the conditions objects, etc. here). But I suspect it's because it's not clear whether satellites need to be included in the model_observatory or not?

yoachim:
Yeah, since this is one potential way of dodging satellites (and not really worth it imo), I figured it was better to not put the updates in the regular `ModelObservatory`. Plus, I like this as an example of how one can add information to the telemetry stream in the conditions object without needing to fork/branch/edit `rubin_sim` code.

yoachim:
Someone should put links/instructions to the conda packaging system(s) in the README file.

PR #226, opened 2022-09-09
compare on disk / expected version info

rhiannonlynne:
I thought it would be convenient to be able to quickly compare the on-disk vs. expected version information.


yoachim (approved):
Sure, as long as this has been tested it looks good.

PR #227, opened 2022-09-09
Tickets/dm 36050

tribeiro:


yoachim (approved):
Looks good. I'll try to go in later today and add a unit test that runs a full scheduler at different nsides. I suspect we also need to update the sky brightness to output at arbitrary nside.

PR #228, opened 2022-09-09
test higher nside works

yoachim:
Add a unit test to make sure we can run at arbitrary nside.

### DIFF
--- a/rubin_sim/skybrightness_pre/dark_sky.py
+++ b/rubin_sim/skybrightness_pre/dark_sky.py
@@ -9,13 +9,13 @@ from rubin_sim.data import get_data_dir
 __all__ = ["dark_sky"]
 
 
-def dark_sky(nside=None):
+def dark_sky(nside=32):
     """Load an array of HEALpix maps that have the darkest expected sky
     backgrounds per filter.
 
     Parameters
     ----------
-    nside : `int`, optional
+    nside : `int` (32)
         Desired nside resolution (default=32).
 
     Returns
@@ -30,14 +30,11 @@ def dark_sky(nside=None):
         dark_sky.data = data["dark_maps"].copy()
         data.close()
 
-    if nside is not None:
-        dark_sky_data = np.empty(hp.nside2npix(nside), dtype=dark_sky.data.dtype)
+    dark_sky_data = np.empty(hp.nside2npix(nside), dtype=dark_sky.data.dtype)
 
-        for band in dark_sky_data.dtype.names:
-            dark_sky_data[band] = hp.pixelfunc.ud_grade(
-                dark_sky.data[band], nside_out=nside
-            )
-    else:
-        dark_sky_data = dark_sky.data
-
+    for band in dark_sky_data.dtype.names:
+        dark_sky_data[band] = hp.pixelfunc.ud_grade(
+            dark_sky.data[band], nside_out=nside
+        )
+    
     return dark_sky_data
--- a/tests/scheduler/test_Baseline.py
+++ b/tests/scheduler/test_Baseline.py
@@ -206,6 +206,37 @@ class TestFeatures(unittest.TestCase):
         # Make sure nothing tried to look through the earth
         assert np.min(observations["alt"]) > 0
 
+    def TestNside(self):
+        """
+        test running at higher nside
+        """
+        nside = 64
+        survey_length = 2.0  # days
+
+        surveys = []
+        # Set up the DD
+        dd_surveys = generate_dd_surveys(nside=nside)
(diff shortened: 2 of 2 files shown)


rubin_sim/skybrightness_pre/dark_sky.py
```diff
@@ -30,14 +30,11 @@ def dark_sky(nside=None):
         dark_sky.data = data["dark_maps"].copy()
         data.close()
 
-    if nside is not None:
-        dark_sky_data = np.empty(hp.nside2npix(nside), dtype=dark_sky.data.dtype)
+    dark_sky_data = np.empty(hp.nside2npix(nside), dtype=dark_sky.data.dtype)
```
tribeiro, line 33: I originally implemented the method this way, but I thought it would be nice to skip the allocation and call to `ud_grade` if `nside=None`. I also think `nside=None` is used as default in other parts of the code so would maintain the same design.

yoachim: I'm kinda tempted to change the the `nside=None` default other places. I figure this function shouldn't get called in any tight loops, so there's no big performance gain to skipping the allocation. 

Also now it should work if someone goes and changes the underlying saved data to a different nside.

tribeiro: Yeah, I thought about that possibility as well... So maybe instead of adding a hard coded value here, we should create a `NSIDE_DEFAULT=32` constant somewhere we can use everywhere in the code. This way we will have an easy way to change it in the future. So maybe add a `constants` submodule?

tribeiro: Also, since you are not that worries about skipping the computation. How about changing the method so it is not skipping reading the file as well (and creating that "data" variable)?

yoachim: We have `utils.set_default_nside` that was intended to be the one spot where you could change the default everywhere all at once. I'd kinda like to get rid of that so there's no auto-magical setting of the nside. It's a trade off, either users need to set lots of `nside=nside` kwargs when they build a scheduler, or we have lots of `if nside is None` blocks in the code. Having `nside=32` defaults all over the place strikes me as slightly more explicit and pythonic, and maybe easier for new folks coming to the code so they don't have to figure out how nside gets magically set.

If we went to super high resolution it would make sense to take out the `data` attribute cache, but it's only a 500K file.

tribeiro: Yeah, I am not a big fan of the `utils.set_default_nside`.. That is not what I am proposing we do. I also think that filling the code with `nside=32` is a maintenance nightmare. 

My proposal is to add a `constants` submodule, where you set `NSIDE_DEFAULT=32`. Then you use `nside=NSIDE_DEFAULT`, and we will have a single place to change it...


But, we might also consider getting rid of this default value altogether. I think this is such an important parameter we might want to force passing this value. At least on the lower levels. Maybe the higher levels that are more user facing could use the default.

tribeiro: > If we went to super high resolution it would make sense to take out the data attribute cache, but it's only a 500K file.

The problem I see is that you are creating this local persistent variable. I don't think we gain anything by doing this. I think you can simply read the file at every call (after all, it is only ~500K).

If this method was called several times at runtime I would understand the need to do this, but in this case, I would actually advocate a different approach. Probably have this as part of the conditions class or something on those line. In fact, it might be a good idea to do that in any case, and not call `dark_sky` in the basis functions and features.

yoachim: OK, I think I like the `constants` submodule solution. 

Moving the dark sky to the conditions object seems like a good idea. It is the dark sky for the site anyway, so it makes sense for it to be in there.

If the current solution works, I'd say we merge it and I can do those changes while I'm refactoring names.

rhiannonlynne:
This looks reasonable to me. 
I would suggest just making one of the changes (reading at every call) now, since it is small and easy to forget. 
I can see how adding the constants module is likely easier to wait, and seems less likely to be forgotten as we'll likely add other items to the constants too.

tribeiro (approved):
Ok, so let's merge this and make the other improvements later. 

Please, note that in the docstring the original text is the standard one 

"`nside : `int`, optional`" instead  of "`nside : `int` (32)`"


See https://numpydoc.readthedocs.io/en/latest/format.html#parameters

yoachim:
Yes, I'll make a note that another part of the refactor will be to standardize the docstring formatting, since we have a few different ones running around.

PR #233, opened 2022-09-14
Verify JSON output works with python 3

rhiannonlynne:
JSON output now works again in python 3. 

```
from io import StringIO
import rubin_sim.maf as maf

b = maf.MetricBundle.load(<some metric file>)
strio  = b.outputJSON()
header, [list of slice points and metric values] = json.loads(strio.getvalue())
```
Note that this output does not currently include masked values. 
It's possible that for work such as @ehneilsen does with schedview, that outputting masked values from the metric with a defined masked value would be more useful than the current approach (which outputs the 'bin' plus the metric value, so skipped binned = masked).


ehneilsen (approved):
Looks fine to me.

PR #236, opened 2022-09-16
U/yoachim/clouds db set

yoachim:
Make it possible to pass in different cloud settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,10 @@ Scheduler, survey strategy analysis, and other simulation tools for Rubin Observ
 [![Run Tests and Build Documentation](https://github.com/lsst/rubin_sim/actions/workflows/python-tests-doc.yml/badge.svg)](https://github.com/lsst/rubin_sim/actions/workflows/python-tests-doc.yml)
 
 
+[![DOI](https://zenodo.org/badge/365031715.svg)](https://zenodo.org/badge/latestdoi/365031715)
+
+
+
 # Installation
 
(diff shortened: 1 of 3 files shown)


rhiannonlynne:
The zenodo citation needs fixing, if this is to be used. 
Please add Scott Daniel to the zenodo author list and also add Tiago's last name (Ribeiro). My name should probably be R. Lynne Jones. 

Best to expand Philippe Gris, Natasha Abrams, and lmptc. It is likely worth considering if all of the contributors should indeed be listed as authors or not, especially as we'd be missing people who contributed to mafContrib and then were merged into rubin_sim and also some who have not been marked as contributors because their contribution came in through a branch and then one of us merged the branch to main (such as Rachel Street or Humna Awan), but I am not certain on the guidelines for citing contributors.  

You might also add some of the rubin community keywords - have a look for other LSST zenodo citables, probably don't join the DM community, but do look at pre-ops related items.

yoachim:
OK, cleaned up the author list and just threw in everyone, because why not. Hopefully zenodo doesn't automatically regenerate the author list every release.

rhiannonlynne:
Does Zenodo auto-build the author list? 
It doesn't look like it necessarily "stuck", as the list looks the same to me as before, but I don't know if it's something that is part of an automatic build chain.

yoachim:
The save button was not enough, had to also hit the publish button apparently.

rhiannonlynne (approved):

PR #237, opened 2022-09-21
ENH first pass at total exposure time metric for riz bands

beckermr:
This PR has a draft metric for the total exposure time in the riz bands after some minimal quality cuts. It is similar to the weak lensing number of visits, but has slightly different constraints.

### DIFF
--- a/rubin_sim/maf/metricBundles/metricBundleGroup.py
+++ b/rubin_sim/maf/metricBundles/metricBundleGroup.py
@@ -1,10 +1,12 @@
 from __future__ import print_function
 from builtins import object
 import os
+import sys
 import numpy as np
 import numpy.ma as ma
 import matplotlib.pyplot as plt
 from collections import OrderedDict
+import tqdm
 
 import rubin_sim.maf.utils as utils
 from rubin_sim.maf.plots import PlotHandler
@@ -447,7 +449,17 @@ class MetricBundleGroup(object):
         else:
             cache = False
         # Run through all slicepoints and calculate metrics.
-        for slice_i in slicer:
+        if self.verbose:
+            slicer_iter = tqdm.tqdm(
+                slicer,
+                desc="Processing slices",
+                ncols=79,
+                file=sys.stdout,
+            )
+        else:
+            slicer_iter = slicer
+
+        for slice_i in slicer_iter:
             i = slice_i["slicePoint"]["sid"]
             slicedata = self.simData[slice_i["idxs"]]
             if len(slicedata) == 0:
--- a/rubin_sim/maf/metrics/vectorMetrics.py
+++ b/rubin_sim/maf/metrics/vectorMetrics.py
@@ -26,8 +26,13 @@ class VectorMetric(BaseMetric):
         metricDtype=float,
         **kwargs
     ):
+        if isinstance(col, str):
+            cols = [col, binCol]
+        else:
+            cols = list(col) + [binCol]
+
         super(VectorMetric, self).__init__(
-            col=[col, binCol], units=units, metricDtype=metricDtype, **kwargs
+            col=cols, units=units, metricDtype=metricDtype, **kwargs
         )
         self.bins = bins
         self.binCol = binCol
--- a/rubin_sim/maf/metrics/weakLensingSystematicsMetric.py
+++ b/rubin_sim/maf/metrics/weakLensingSystematicsMetric.py
@@ -133,7 +133,7 @@ class RIZDetectionCoaddExposureTime(VectorMetric):
     ugrizY are present. We do not make a depth cut explicitly since that is
     circular (and thus confuses MRB's feeble mind :/).
 
-    TODO:
+    TODO maybe:
      - apply some sort of inverse variance weighting to the coadd based on sky
        level?
      - use some sort of effective exposure time that accounts for the PSF?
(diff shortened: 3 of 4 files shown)


rhiannonlynne:
Could you please run 'black' on the file? I think the setup you're looking for is the same as the one that DM uses, but in case it's not, the requirements-test.txt file covers it I think.

beckermr:
Comments are great! I plan to run black I just didn't get to it. :)

rhiannonlynne (commented):
I get a little confused sometimes about where GitHub sends my comments. I made two comments in line in the text (one just a question, one that may or may not be helpful) and I see from my end that they are "Pending" which makes me wonder if they are visible or not. So I'm going to hit "Submit Review" as I think that makes them stop being pending and makes sure they're visible.

rubin_sim/maf/metrics/weakLensingSystematicsMetric.py
```diff
@@ -116,3 +116,87 @@ def run(self, dataSlice, slicePoint):
...
+            return self.badval
+
+        # check to make sure there is at least some
+        # coverage in the required bands
+        filters = set(dataSlice[self.filterCol])
+        if filters != self.min_bands:
+            return self.badval
+
+        exptime_msk = (dataSlice[self.expTimeCol] > self.min_expTime)
+        filter_msk = None
```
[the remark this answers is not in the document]
beckermr: That'd be fine too I think. Probably better since it makes fewer arrays! Thanks!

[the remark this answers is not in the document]
beckermr: I moved to a different function to do this.

rubin_sim/maf/metrics/weakLensingSystematicsMetric.py
```diff
@@ -116,3 +116,87 @@ def run(self, dataSlice, slicePoint):
...
+    metadetection (weak lensing shear estimator) and will likely be adopted by the Rubin science
+    pipelines.
+
+    It counts the total exposure time in all three bands, excluding dusty regions, exposures
+    that are too short, or areas where not all bands ugrizY are present. We do not make a depth
+    cut explicitly since that is circular (and thus confuses my feeble mind :/).
+
+    TODO:
+     - apply some sort of inverse variance weighting to the coadd based on sky level?
+     - use some sort of effective exposure time that accounts for the PSF?
```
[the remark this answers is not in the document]
beckermr: Right. The final weight of each image in the cell-based coadd will be something like an inverse variance weight based on the sky background pixel noise. If we move to Kaiser coadds, this weight will get even more complicated. I'm thinking now that instead of dealing with all of this uncertainty, it is simpler to count raw exposure time with the expectation that some effects go one way (bad seeing makes it worse) and others go the other (better weighting in the coadding reduces the effects of bad exposures).

rubin_sim/maf/metricBundles/metricBundleGroup.py
```diff
@@ -447,7 +449,17 @@ def _runCompatible(self, compatibleList):
...
+            slicer_iter = tqdm.tqdm(
+                slicer,
+                desc="Processing slices",
+                ncols=79,
+                file=sys.stdout,
+            )
+        else:
+            slicer_iter = slicer
+
+        for slice_i in slicer_iter:
```
beckermr, line 462: I can back this out if you want. It is very nice to know when the metric will finish.

rubin_sim/maf/metrics/vectorMetrics.py
```diff
@@ -26,8 +26,13 @@ def __init__(
...
         **kwargs
     ):
+        if isinstance(col, str):
+            cols = [col, binCol]
+        else:
+            cols = list(col) + [binCol]
+
         super(VectorMetric, self).__init__(
-            col=[col, binCol], units=units, metricDtype=metricDtype, **kwargs
+            col=cols, units=units, metricDtype=metricDtype, **kwargs
```
beckermr, line 35: This appears to be a bug or something?

rubin_sim/maf/metricBundles/metricBundleGroup.py
rhiannonlynne: Looks like it's pretty minor addition, only gets triggered with "verbose" which shouldn't happen very often, and is a minor time increase in execution when it does. 
We can add it to the requirements.txt file and update the conda feedstock to add the package. 
(do you want to add it to the requirements.txt file here, so that the GitHub actions pass?)

rhiannonlynne (approved):

rubin_sim/maf/metrics/vectorMetrics.py
beckermr: The old version did not work for me yes.

beckermr:
@rhiannonlynne This is ready to merge if you are happy with it!

rhiannonlynne:
I'm not sure why the documentation upload keeps failing for this branch. It doesn't seem to be anything specific here, just a failure to upload to lsst.io, which does happen sometimes but this seems more often. I just reran that part of the job to see if it will work this time, and if not, can check with @jonathansick. 

The Jenkins failure is in the lsst-ts Jenkins builds. They're using their own lsst-ts conda setup, and tdqm will have to be added to their packages and then their Jenkins setup restarted (I think). I can send Tiago a PR on that, just to avoid a few small hoops. 
Or - if you felt like having a look at their setup, it's in https://github.com/lsst-ts/ts_recipes

tdqm will have to be added to the rubin-sim-feedstock on condo-forge as well, if you feel like sending that PR, otherwise I guess the way to do it is to merge this, update Rubin-sim-feedstock, then tag here so that the conda build gets updated? (I'm not entirely sure of the best procedure).

rhiannonlynne:
I'll go ahead and merge.

PR #238, opened 2022-09-22
U/lynnej/update actions

rhiannonlynne:
Update download link from USDF
Update which actions run on release creation and update python version to 3.10


yoachim (approved):
That looks in order.

PR #242, opened 2022-10-17
U/lynne/last minute metric changes

rhiannonlynne:
Fix gamma file for DDF SNIa and DDF field names
In AGN SF uncertainty, add more bins to SF error for DDFs.

### DIFF
--- a/rubin_sim/data/rs_download_data.py
+++ b/rubin_sim/data/rs_download_data.py
@@ -25,7 +25,7 @@ def data_dict():
         "skybrightness": "skybrightness_may_2021.tgz",
         "skybrightness_pre": "skybrightness_pre_2022_5_18.tgz",
         "throughputs": "throughputs_aug_2021.tgz",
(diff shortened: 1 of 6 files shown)


yoachim:
Is there a `test_simData.hdf` file running around somewhere? Looks like I need that to run the unit tests. Is `main` even passing unit tests right now?

rhiannonlynne:
It is - but it's because we don't run the full set of unit tests (and since this requires the MAF data to be present, and that's large enough that it's slow, we don't run those all the time). 
So thanks! 
I should have added an updated 'tests' data file, which I can't put onto USDF because I don't have permissions. 
Can you take the tests data file from https://epyc.astro.washington.edu/~lynnej/opsim_downloads/rubin_sim_data/tests_2022_10_18.tgz
and upload it at USDF for me please? 
(this adds that tests_simData.hdf5 file to the tests data directory). 
I updated the version info accordingly.

yoachim:
commits merged in from a different PR.

PR #243, opened 2022-10-18
U/yoachim/ddf metric simplify

yoachim:
Update SNe metric so it doesn't try to do inappropriate selections that should be done in the sql query. Get rid of pandas future warning.

### DIFF
--- a/rubin_sim/maf/batches/ddfBatch.py
+++ b/rubin_sim/maf/batches/ddfBatch.py
@@ -39,6 +39,7 @@ def ddfBatch(
     # Get standard DDF locations and reformat information as a dictionary
     ddfs = {}
     ddfs_rough = ddf_locations()
+    ddf_sn_fieldnames = [f"DD:{k}" for k in ddfs_rough.keys()]
     for ddf in ddfs_rough:
         ddfs[ddf] = {"ra": ddfs_rough[ddf][0], "dec": ddfs_rough[ddf][1]}
     # Combine the Euclid double-field into one - but with two ra/dec values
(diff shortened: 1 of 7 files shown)


rubin_sim/maf/batches/ddfBatch.py
```diff
@@ -130,9 +125,7 @@ def ddfBatch(
             zStep=0.03,
             daymaxStep=3,
             coadd_night=True,
-            gammaName=gammaName,
-            fieldType=fieldType,
-            DD_list=[ddf_sn_fieldnames],
+            gammaName="gamma_DDF.hdf5",
```
rhiannonlynne, line 127: Seems more appropriate to use the WFD gamma file for the WFD field though?

rubin_sim/maf/batches/scienceRadarBatch.py
```diff
@@ -732,10 +732,11 @@ def scienceRadarBatch(
         verbose=False,
     )
     plotDict = {"percentileClip": 95, "nTicks": 5}
+    # Run without DDF observations
     bundle = mb.MetricBundle(
         metric,
         snslicer,
-        "",
+        "note not like %DD%",
```
rhiannonlynne: Can you humor me and add the info_label as well, so that it says more simply "non-DD"  in the plot title/metric file, etc. instead of using this awkward sql constraint in those places?

rubin_sim/maf/batches/ddfBatch.py
rhiannonlynne: Also - is a sqlconstraint missing here to exclude WFD visits from the bundle? 
(it does make more sense to me to use the WFD visits as well, as it helps fill in epochs, etc., but Philippe set it up this way purposefully and if we want to continue to have consistent results, we'd need to exclude those as well).

tests/maf/test_SNmetrics_NSN.py
```diff
@@ -58,7 +59,9 @@ def testNSN(self):
             [(5.158324, 0.276509)], names=["nSN", "zlim"]
         )
         # These are of the same point on the sky, with/without DD fields - they should be the same
-        # as the metric rejects DD visits internally
+        # as the metric rejects DD visits internally---XXX except now the metric does
+        # not reject visits based on notes, so not sure why these still pass fine.
```
rhiannonlynne, line 63: Not sure - worth poking at. I wonder if it's the gamma file.

rhiannonlynne (approved):

rubin_sim/maf/batches/ddfBatch.py
yoachim: The point of picking a random spot in the WFD and treating it like a DDF was to sanity check all the DDF metrics. Using the WFD gamma file we got the result that COSMOS would get 13 SNe and a random spot in the WFD would have 80. 

Sure, we can exclude non-ddf observations.

PR #246, opened 2022-11-02
Fix young metric

Thalos12:
This change is necessary to take into account the effect of crowding.

Main changes:
- use `rubin_sim.maf.metrics.crowdingMetric.CrowdingM5Metric` to get crowding depth
- take depth as minimum between m5 and crowding depths
- optionally return distance corresponding to depth if the `returnDistance` keyword argument is set to True

### DIFF
--- a/rubin_sim/maf/mafContrib/YoungStellarObjectsMetric.py
+++ b/rubin_sim/maf/mafContrib/YoungStellarObjectsMetric.py
@@ -72,6 +72,11 @@ class NYoungStarsMetric(BaseMetric):
         The galactic latitude above which to return zero (degrees).
     badval : float, opt
         The value to return when the metric value cannot be calculated. Default 0.
+    
+    Keyword arguments
+    -----------------
+    returnDistance : bool, opt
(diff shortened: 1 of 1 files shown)


rhiannonlynne (commented):
This looks reasonable, and I'll double check that it works for us shortly. 
Given the timing with the SCOC review, I don't think we can get this in place for that evaluation unfortunately. (rerunning on all of the older simulations would be intensive). 
However, we can certainly run a few of the newer simulations with this new version to look at how much it changes the results.

Thalos12:
It is my fault, I made the pull request later than I intended to, but I think it would be great if the metric could be run on the newer simulations at least. Thank you!

rhiannonlynne (approved):

rhiannonlynne:
Approve, but I guess we should merge this to the new dev-camel-snake branch (after appropriate updates) instead. 
Also, the file changed needs "black" to be run on it, as it's failing tests due to that.

Thalos12:
Thank you very much for merging the pull request.

Apologies for not running black on the code, I will remember to do it before a making a pull request next time.

PR #249, opened 2022-11-23
Tickets/dm 36852

tribeiro:


rubin_sim/scheduler/features/features.py
```diff
@@ -515,12 +515,17 @@ class NoteLastObserved(BaseSurveyFeature):
...
         self.note = note
+        self.filtername = filtername
         self.feature = None
 
     def add_observation(self, observation, indx=None):
-        if self.note in observation["note"]:
+        if self.note in observation["note"] and (
+            self.filtername == observation["filter"]
+            if self.filtername is not None
+            else True
```
rhiannonlynne, line 527: That seems like some convoluted-to-read python for a late-at-night at the observatory person. I know it's the same, but is there actually a performance penalty for making it an if/else inside the first "if"? 
ie.. more like 
```
if self.note in observation['note']:  
   if (self.filtername is None) or (self.filtername == observation['filter']):
       self.feature = observation['mjd']
```
(should probably note in the doc string that observation note matches if self.note is contained in the note, not just exact matches .. which 'substring' may be intended to convey, but might not entirely?)

rubin_sim/scheduler/basis_functions/basis_functions.py
```diff
@@ -1907,30 +1907,46 @@ class VisitGap(Base_basis_function):
     ----------
     note : str
         Value of the observation "note" field to be masked.
+    filter_names : list [str], optional
+         List of filter names that need be observed before activating.
```
rhiannonlynne, line 1911: "Before activating" -- would it also make sense to reword and extend this as 
'List of filter names that will be considered when evaluating if the `gap` has passed. All filters must be observed before the gap requirement will be activated, and once activated, only observations in these filters will be evaluated in context of whether the last observation was at least `gap` in the past.'

tests/scheduler/test_BasisFuncs.py
```diff
@@ -40,6 +40,102 @@ def test_label(self):
...
+
+        # observation with the correct note but only one filter
+        assert visit_gap.check_feasibility(conditions=conditions)
+
+        observation["filter"] = "i"
+        visit_gap.add_observation(observation=observation)
+
+        # observation with the correct note and both filters
+        assert not visit_gap.check_feasibility(conditions=conditions)
+
```
rhiannonlynne, line 138: I would suggest adding a test to check that the visit gap turns back to 'feasible' after the time period has passed -- especially when multiple filters are involved, in case behavior ends up changing (you may want to test in particular, adding enough observations to trigger the "gap" requirement, then adding more observations in both a non-contributing bandpass and a bandpass in filterlist, then checking if those observations at a later time are feasible (in particular if you add an r band and then immediately a g band).

rhiannonlynne:
Looks like you probably need this extension because the aux tel surveys you're using have multiple filters (or share a basis function between surveys)?

rhiannonlynne (approved):
Minor comments.

rubin_sim/scheduler/features/features.py
tribeiro: good point.. 

I merged your second if into the first, which gives the same outcome:

```
        if self.note in observation["note"] and (
            self.filtername is None or self.filtername == observation["filter"]
        ):

```

rubin_sim/scheduler/basis_functions/basis_functions.py
tribeiro: I kept your first phrase in the parameter docstring and moved everything else to a Notes section.

PR #252, opened 2022-12-05
U/yoachim/hg notebook fix

yoachim:
A few lingering things from the 1.0 update that only showed up when running notebooks.


ehneilsen (approved):
Okay.

PR #253, opened 2022-12-07
replace .asscalar with cast to float

ehneilsen:
The rubin_sim Conditions._str_ implementation currently uses the .asscalar method of numpy array, which has been dropped in newer versions of numpy. Replace this call so that it works with more recent versions of numpy.


yoachim (approved):
yup

PR #254, opened 2022-12-07
Tickets/preops 3129

ehneilsen:
Change names in Conditions.__str__ to match what is used in 1.0.0 (and onward)


yoachim (approved):
Looks good. Might be time to add a unit test since I don't think `__repr__` and `__str__` are getting called in any of the other tests.

ehneilsen:
> Looks good. Might be time to add a unit test since I don't think `__repr__` and `__str__` are getting called in any of the other tests.

Yes, that's a good idea. I'll work on that, too.

PR #255, opened 2022-12-08
Update LST in Conditions when necessary

ehneilsen:
When the MJD for a conditions object is assigned, _lmst in set to None because the old value is on longer valid. Have the getter compute it anew when needed, rather than returning None.


yoachim (approved):
Looks good

PR #260, opened 2023-01-06
Add area and accum area to reward table

ehneilsen:
Expands methods that produce DataFrames of rewards to include unmasked area, where applicable.  I expect to make further modifications to refine the contents of this DataFrame, but what is here now should get merged now so it doesn't stop progress on moving visualization elements from schedview to rubin_sim (and it doesn't actually make anything worse than it already is).


yoachim (approved):
That all looks good. It'd be good if there's at least a little bit of unit test coverage (I think there might be, but I'm too lazy to check).

ehneilsen:
There was some coverage of my new code, but not, I think, enough, so I added a new test.

PR #262, opened 2023-01-15
U/lynnej/backwards compatibility

rhiannonlynne:
Backward compatibility for metric files and resultsDB files. 
Also updated the summary stat grab so it's much faster.
Updated show_maf pages to have correct web docs and removed (unused) config links.

### DIFF
--- a/rubin_sim/maf/run_comparison/archive.py
+++ b/rubin_sim/maf/run_comparison/archive.py
@@ -41,7 +41,7 @@ METRIC_SET_SOURCE = os.environ.get(
 
 SUMMARY_SOURCE = os.environ.get(
     "RUBIN_SIM_SUMMARY_SOURCE",
-    "https://raw.githubusercontent.com/lsst-pst/survey_strategy/main/fbs_2.0/summary_2022_11_16.csv",
+    "https://raw.githubusercontent.com/lsst-pst/survey_strategy/main/fbs_2.0/summary_2023_01_01.csv",
 )
 
 if os.uname().nodename.endswith(".datalab.noao.edu"):
--- a/rubin_sim/maf/db/add_run.py
+++ b/rubin_sim/maf/db/add_run.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import argparse
+from . import ResultsDb
 from . import add_run_to_database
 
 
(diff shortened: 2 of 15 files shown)


rubin_sim/maf/run_comparison/archive.py
```diff
@@ -41,7 +41,7 @@
 
 SUMMARY_SOURCE = os.environ.get(
     "RUBIN_SIM_SUMMARY_SOURCE",
-    "https://raw.githubusercontent.com/lsst-pst/survey_strategy/main/fbs_2.0/summary_2022_11_16.csv",
+    "https://raw.githubusercontent.com/lsst-pst/survey_strategy/main/fbs_2.0/summary_2023_01_01.csv",
```
yoachim, line 44: Should this be smart enough to grab the most recent version in case it changes? Or just put the most recent copy in `rubin_sim_data`?

yoachim (approved):
Seems like `gather_summaries` still doesn't scale that well. Not really urgent, I can add my faster script later.

rubin_sim/maf/run_comparison/archive.py
rhiannonlynne: Most likely, yes that is something we should do instead. I'm ok with leaving it like this for now .. especially as we probably want to figure out a better versioning setup.

rhiannonlynne:
When you say it doesn't scale that well - can you expand on that? 
Running on a single resultsDb, there should be something like a 1000 times speedup, based on %%timeit tests in a notebook cell. But, I didn't really benchmark it on multiple databases with varying metrics, which is where the join/concat part comes back in again, and of course, using sqlalchemy is likely slower than pandas/read_sql.

rhiannonlynne:
(I'm going to merge, but is ok to finish that convo above .. )

yoachim:
It takes over 10 minutes to combine the MAF results from v2.1 (so around 130 sims). The quick and dirty script I've been using can do it in 80 seconds. There's probably a rogue loop that's going over all the metric values when they could all be read at once.

rhiannonlynne:
It should be querying all the metrics at once now, although it does have to loop over the results (because of the dqlalchemy query) to put them in a list to put into the numpy array. If it's doing multiple queries per database, that should be a bug.

PR #263, opened 2023-01-17
U/yoachim/slicer needed cols

yoachim:
Minor bug fixes:
* make sure MAF maps can be set as strings or instantiated maps or mix of the two
* spatial slicer needs the rotation column name even if not using the chip gaps
* Let `sim_runner` complete even if it took no observations


ehneilsen (approved):
Looks reasonable.

PR #267, opened 2023-01-20
U/yoachim/fast gather summaries

yoachim:
Update gather_summaries so it can scale hundreds of simulations.


rhiannonlynne:
I'm pretty sure RunComparison is being used in at least a few notebooks out there, including in our own tutorials.

yoachim:
Looks like `RunComparison` is only used in `GalacticPlaneMetrics.ipynb` which hasn't been updated to 1.0 anyway.

rubin_sim/maf/run_comparison/gather_summaries.py
```diff
@@ -1,16 +1,99 @@
...
+    sql_q += "summarystats.summary_value "
+    sql_q += "FROM summarystats INNER JOIN metrics ON metrics.metric_id=summarystats.metric_id"
+
+    rows = []
+    for row_name, fname in zip(run_names, db_files):
+        con = sqlite3.connect(fname)
+        temp_df = pd.read_sql(sql_q, con)
+        con.close()
+
+        # Use maf.db.ResultDb.build_summary_name so it matches show_maf labels
```
rhiannonlynne, line 53: Oh ha. I really thought that the contents on the show maf pages matched this, but when I double checked, they don't. 
So forget about it and just do whatever thing you were doing before.
Eventually we'll update all of the downstream stuff and dashboards anyway.

rubin_sim/maf/run_comparison/gather_summaries.py
```diff
@@ -47,33 +123,19 @@ def gather_summaries():
     )
     args = parser.parse_args()
 
-    # Identify a subset of MAF directories if suffix is set
-    if args.suffix is None:
-        run_dirs = None
-    else:
-        run_dirs = glob.glob(f"*{args.suffix}")
+    run_dirs = glob.glob(args.base_dir + "/*/")
```
rhiannonlynne, line 126: So, it would be nice to be able to choose a particular set of directories, but I know you don't run your workflow that way.  I guess if we want to access just a subset, we could run the "gs" function above (that's not a super descriptive name, but ok).

rhiannonlynne (approved):

rhiannonlynne:
> Looks like `RunComparison` is only used in `GalacticPlaneMetrics.ipynb` which hasn't been updated to 1.0 anyway.

That works then. 


It kind of seems like the entire "build_summary_name" method could be dropped from ResultsDb too? 
If the new summary name is just summary name + metric name + info label + slicer name  (note that you do need to preserve the slicer information, otherwise healpix slicers vs. one d slicers vs. unislicers are not easy to tell apart). 
I think you previously had the summary name last, but this seems like it would be odd. 
So, we would get names like
"identity median m5 g band unislicer" 
"median mean m5 g band healpix slicer" 
which seems clear enough to me.

PR #269, opened 2023-01-23
Fix ra/dec unit conversion bug in FieldSurvey class.

edennihy:


rubin_sim/scheduler/surveys/field_survey.py
```diff
@@ -48,13 +48,13 @@ def __init__(
             detailers=detailers,
         )
         self.basis_weights = np.ones(len(basis_functions)) / len(basis_functions)
+        self.ra_deg, self.dec_deg = np.rad2deg(self.ra), np.rad2deg(self.dec)
```
tribeiro, line 51: Picky comment, but I think you can avoid the double conversion (from deg -> radians -> deg) by using `RA` and dec, e.g.;

```
self.ra_deg = RA
self.dec_deg = dec
```

tribeiro (approved):
I left a small inline comment, please take a look before merging.. Otherwise, looks good..

PR #270, opened 2023-01-26
make sure conditions.cumulative_azimuth_rad is set

yoachim:
Looks like cumulative_azimuth_rad was added to the docstring but never made it into the code.


tribeiro (approved):
looks good!

PR #271, opened 2023-01-26
remove hp.UNSEEN values

yoachim:
Remove hp.UNSEEN since it's technically a float that can behave strangely so `np.nan` usually makes more sense.

### DIFF
--- a/rubin_sim/maf/metrics/summary_metrics.py
+++ b/rubin_sim/maf/metrics/summary_metrics.py
@@ -199,7 +199,7 @@ class TotalPowerMetric(BaseMetric):
         lmin=100.0,
         lmax=300.0,
         remove_dipole=True,
-        mask_val=hp.UNSEEN,
+        mask_val=np.nan,
         **kwargs
     ):
         self.lmin = lmin
--- a/rubin_sim/maf/plots/spatial_plotters.py
+++ b/rubin_sim/maf/plots/spatial_plotters.py
@@ -326,11 +326,11 @@ class HealpixPowerSpectrum(BasePlotter):
             return None
         if plot_dict["removeDipole"]:
             cl = hp.anafast(
-                hp.remove_dipole(metric_value.filled(slicer.badval)),
+                hp.remove_dipole(metric_value.filled(hp.UNSEEN)),
                 lmax=plot_dict["maxl"],
             )
         else:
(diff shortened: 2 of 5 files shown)


rhiannonlynne:
This isn't a bad idea - it's true that adding together two healpix grid metrics while just using "filled" will give you what would-have-been-masked pixels that end up not being masked. (I assume that's the behave strangely you're talking about? or is it something else?) 
But I'm surprised you don't have to now change something in the plotting methods so that the masked values still look blank? Or is this happening by accident because of the color bar remapping that's done in those methods? 
Also - if we're going to use np.nan .. does it make sense to go a step further and abolish the masked array completely and just use np.NaN where we want things to be masked? (and then add appropriate work-around-the-nan in places where we'd been using the mask).

rhiannonlynne (approved):

rhiannonlynne:
Oh - actually --- I just did a test and I think in that first file you do need to then modify the metric to set the fill value to hp.UNSEEN when calling anafast. 
Before, when we were using hp.UNSEEN as the mask fill value, anafast would work without having to specify this. But if we swap to np.nan as the fill value, then we should specify hp.UNSEEN for the fill value when calling anafast. 

Screenshot 2023-01-27 at 8 34 48 AM

rhiannonlynne:
Oh right - and this probably shows up when calling the power spectrum plotting too (in plots.HealpixPowerSpectrum).

rubin_sim/maf/metrics/summary_metrics.py
```diff
@@ -199,7 +199,7 @@ def __init__(
         lmin=100.0,
         lmax=300.0,
         remove_dipole=True,
-        mask_val=hp.UNSEEN,
+        mask_val=np.nan,
```
rhiannonlynne, line 202: Later in this metric though, I think you might need to do something like 
data = np.where(np.isnan(data_slice[self.col_name]), hp.UNSEEN, data_slice[self.col_name])
and then use data in the hp.anafast call. 
(since data_slice[self.col_name] here is actually the metric values, which will have been filled with the mask value). 
I thought at some point we were calculating the "TotalPower" summary metric, for desc? But maybe it's not been used for a while.

rhiannonlynne (approved):

PR #272, opened 2023-01-27
Slim down moving object discovery metric output

rhiannonlynne:
Make moving object discovery metrics only report the first discovery instead of all of the potential discovery chances. 
This removes some potential capability, but in practice we haven't used these other options.


yoachim (approved):
That all seems to be in order.

PR #273, opened 2023-01-30
sqlalchemy 2.0 requires 'text' for queries

rhiannonlynne:
This fixes the failure that happens in sqlalchemy 2.0, when a plain string is passed as a query. (sqlalchemy enforces queries to be sqlalchemy.sql.text).


tribeiro (approved):
Looks good!

PR #275, opened 2023-02-02
Update to black v23 (from v22). 

rhiannonlynne:
.. also added de421.bsp to the .gitignore file.


yoachim (approved):
Looks like black decided it doesn't like blank lines. Sure, works for me.

PR #276, opened 2023-02-07
U/yoachim/presto mjd0

yoachim:
Two science metrics need the survey start explicitly set, so added that so it's easy to compare runs with different start dates. Also fixed a jinja formatting issue so it doesn't crash on NaNs.


rhiannonlynne (approved):
Looks reasonable. 
Are there other metrics that use mjd0? Might only be solar system metrics?

PR #277, opened 2023-02-08
support field surveys with rewards for scalar basis functions

ehneilsen:
Auxtel users encountered a problem in schedview caused by scalar basis functions in the field surveys. This fix lets then use either scalar or array basis functions.


tribeiro (approved):
This looks good. Could add a unit test for this in https://github.com/lsst/rubin_sim/blob/13a42349cf77b3f40e20b09e7033870fdf552145/tests/scheduler/test_surveys.py#L12?

PR #278, opened 2023-02-10
Update young stars metric

Thalos12:
Two main changes:
- allow using 2D extinction instead of 3D one
- allow disabling crowding effects


yoachim:
Merged from other branch

PR #279, opened 2023-02-10
Fix CrowdingMagUncertMetric

Thalos12:
The metric was not using the best value of the seeing, resulting in higher than expected crowding errors.

### DIFF
--- a/rubin_sim/maf/metrics/crowding_metric.py
+++ b/rubin_sim/maf/metrics/crowding_metric.py
@@ -244,7 +244,7 @@ class CrowdingMagUncertMetric(BaseMetric):
         lum_func = slice_point[f"starLumFunc_{self.filtername}"]
         # Magnitude uncertainty given crowding
         dmag_crowd = _comp_crowd_error(
-            mag_vector, lum_func, data_slice[self.seeing_col], single_mag=self.rmag
+            mag_vector, lum_func, min(data_slice[self.seeing_col]), single_mag=self.rmag
         )
         result = np.mean(dmag_crowd)
         return result


yoachim:
There is potential to have an artificially low seeing value if you take the minimum. For example, the best seeing could be in twilight time when the sky background is high. Should it be slightly more sophisticated? Maybe the best seeing value where the 5-sigma limiting depth is above the 50th percentile?

Thalos12:
My intent with this change is to get results that are consistent with those from `CrowdingM5Metric`, as I am under the impression that these two metrics are one the inverse of the other. For instance, in the u band (with UserPointsSlicer(270,-30))
```
CrowdingM5Metric(0.25, "u" , maps=["TrilegalDensityMap"]) ==> 22.8
CrowdingM5UncertMetric(22.8, "u" , maps=["TrilegalDensityMap"]) ==> 0.44
```
while I would have expected a value closer to 0.25 mag. Adding the `min` "fixes" the issue, but I did it because it was missing only in `CrowdingM5UncertMetric` and was present everywhere else. In fact, I did not consider the case you mentioned, and it would make sense to adopt a more realistic value, but then should it be changed also in the other metrics I guess.

rhiannonlynne:
The other problem is that neither of these accounts for the photometric noise, although the docstring indicates that they do. Or am I missing something?  
I do appreciate your point, that the outputs of these should be the same.

rhiannonlynne:
Ah - and I see in your NYoungStars metric, that you have to account for the fact that the crowdingm5metric isn't accounting for photometric depth too ... 
Would modifying the CrowdingM5Metric to return m5 accounting for both photometric depth and crowding limits make sense? I suspect it would be more useful in general.

Thalos12:
> neither of these accounts for the photometric noise

I am not sure if this was made on purpose or not, I am not familiar enough with the way crowding is computed. I will try to read more on it, though, and I'll ask somebody more knowledgeable than me on the topic.

> Would modifying the CrowdingM5Metric to return m5 accounting for both photometric depth and crowding limits make sense

That might be useful, but I'll have to ask first to be sure.

I'll try to give a reply as soon as possible. Thank you for your patience.

PR #284, opened 2023-02-28
U/yoachim/example scheduler

yoachim:
add a baseline-like example scheduler so it's easy to generate a realistic scheduler object for testing, etc. Will update DDF scheduling on later PR.

### DIFF
--- a/rubin_sim/scheduler/example/example_scheduler.py
+++ b/rubin_sim/scheduler/example/example_scheduler.py
@@ -1,0 +1,1475 @@
+import numpy as np
+import matplotlib.pylab as plt
+import healpy as hp
+from rubin_sim.scheduler.schedulers import CoreScheduler
+from rubin_sim.scheduler.utils import (
+    SkyAreaGeneratorGalplane,
+    ConstantFootprint,
+    make_rolling_footprints,
+)
+import rubin_sim.scheduler.basis_functions as bf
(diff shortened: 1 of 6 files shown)


rubin_sim/scheduler/example/example_scheduler.py
```diff
@@ -0,0 +1,1496 @@
...
+        )
+        scripted = ScriptedSurvey([], nside=nside, ignore_obs=["blob", "DDF", "twi"])
+        surveys.append(
+            LongGapSurvey(blob[0], scripted, gap_range=gap_range, avoid_zenith=True)
+        )
+
+    return surveys
+
+
+def gen_GreedySurveys(
```
ehneilsen, line 361: Maybe call this get_GreedySuveys? Using caps suggests to me that it is defining a new class (or list of new classes), rather than creating a list of instances of existing classes.

ehneilsen (approved):
This is awesome, I'm looking forward to using it.

There are a few cases of `# XXX do something` that could perhaps be cleaned up, and one function name I commented on.

Several of the functions have lists of assignments, embedding parametrization in the code. Eventually, it might be nice to be able to read these values from a user specified yaml file or some such instead. (This is just an idea, not really a suggestion for this commit, and not a high priority, or even something I, personally, have any use for just now.)

PR #285, opened 2023-02-28
Raise a more descriptive error message for out of range points

yoachim:
Update so if users try to compute sky values for coordinates that are not visible or below model airmass limits then a error gets raised.


rhiannonlynne (approved):

yoachim:
Looks like SDF is having issues, so merging anyway.

PR #287, opened 2023-03-03
Fix selection of labels to signify inversion in metric/run mesh plots

ehneilsen:
Sometimes, the old code modified the metric labels of the wrong metrics to reflect inversion.

### DIFF
--- a/rubin_sim/maf/run_comparison/summary_plots.py
+++ b/rubin_sim/maf/run_comparison/summary_plots.py
@@ -493,11 +493,30 @@ def plot_run_metric_mesh(
     if metric_label_map is None:
         metric_labels = metrics
     else:
-        metric_labels = [metric_label_map[m] for m in metrics]
-    # Update the plot label if we inverted the column during normalization
-    if baseline_run is not None and metric_set is not None:
-        for i in np.where(metric_set["invert"])[0]:
-            metric_labels[i] = f"1/{metric_labels[i]}"
+        try:
+            # When there are multiple metric families specified,
+            # there might be duplicate elements in the
+            # metric_label_map. Remove the duplicates.
+            metric_label_map = metric_label_map[
+                ~metric_label_map.index.duplicated(keep="first")
+            ]
+        except AttributeError:
+            # if metric_label_map is a dict, it won't have
+            # an index attrubute, but there isn't any danger
+            # if duplicates either
+            pass
+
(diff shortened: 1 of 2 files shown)


ehneilsen:
Don't approve this merge quite yet; now that I fixed the test, I need to do a rebase.

rhiannonlynne:
Ah I see the problem you were solving with the initial PR (about the inversion of metric labels) - more along the lines of if there were multiple metrics with the same name in the metric_set then it was a problem.  I  was curious, as I hadn't seen it go wrong anywhere. 

I had been having the "multiple metrics with the same name" issue, though, and just solving it by removing the duplicates before sending in the 'metric set'. This will be cleaner - nice fix.

rhiannonlynne:
As far as testing for the specific file instead of the directory .. it's a little more brittle in case we change the filenames or default nside (and then removed nside=32) for some reason. However, it is also a more specific test and those are unlikely scenarios, so seems reasonable to me as well.

ehneilsen:
I think something went wrong in a conda dependency. Nothing that changed since the last time it passed seems at all related to the failure, and it’s failed like this twice today.
________________________________

ehneilsen:
I switched to using the specific file rather than the directory, because checking for the directory didn't actually work. It looks like the directory itself is actually created in the tests, so testing for the directory doesn't skip the tests.

From: Lynne Jones ***@***.***>
Date: Tuesday, March 7, 2023 at 2:02 PM
Subject: Re: [lsst/rubin_sim] Fix selection of labels to signify inversion in metric/run mesh plots (PR #287)


As far as testing for the specific file instead of the directory .. it's a little more brittle in case we change the filenames or default nside (and then removed nside=32) for some reason. However, it is also a more specific test and those are unlikely scenarios, so seems reasonable to me as well.

ehneilsen:
The fix wasn't only about dealing with duplicates: in the old way, I was seeing the wrong labels get modified even when there weren't duplicates. Duplicates were also a problem, though.

rhiannonlynne:
I asked about the failure on the #dm-conda channel. 
I think it's a conda-forge migration from jpeg to jpegturbo that's happening and is causing some issues. 

It builds in Jenkins, so I think you can probably just go ahead and merge. This isn't a problem at our end, and unless people are building the environment from scratch, it won't be a problem for them either.

rhiannonlynne (approved):

PR #289, opened 2023-03-08
U/yoachim/fov vectorize

yoachim:
Update the code that checks if a HEALpixels are in a pointing to accept arrays of positions.


rhiannonlynne:
Looks good!  Is this added to help support reloading pointings into the features more rapidly?

rhiannonlynne (approved):

yoachim:
> Looks good! Is this added to help support reloading pointings into the features more rapidly?

Yup. This should be a neutral change now, but will enable reading in observations as a large array rather than one at a time.

PR #290, opened 2023-03-08
U/yoachim/add ddf sched

yoachim:
Add code for pre-computing DDF schedules and add it to the example scheduler.


ehneilsen (approved):
This looks fine

PR #292, opened 2023-03-13
Places I had to add names and tools

rhiannonlynne:
This might be a useful PR for a starting point for "places that didn't have survey_name that now do". 
Still didn't manage to stuff names into the DDF surveys though. 
I also don't think this takes things quite far enough as names aren't put onto the survey_df etc that show up in schedview. 
But still, it might be helpful.

### DIFF
--- a/rubin_sim/scheduler/model_observatory/kinem_model.py
+++ b/rubin_sim/scheduler/model_observatory/kinem_model.py
@@ -659,6 +659,6 @@ class KinemModel(object):
         visit_time = self.visit_time(observation)
         # Compute any overhead that is left over from this
         if ~np.isnan(slewtime):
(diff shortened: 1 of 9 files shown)


yoachim (approved):
All seems fine.

Should we plan to completely decouple `survey_name` and `note` at this point? Maybe add `survey_name` to the `observation` dtype so both get recorded?

rhiannonlynne:
Regarding the question of survey name and survey note -- I kind of think this is a good idea, but we should check that the downstream actual users can accomodate a different "Target". But yes, I think we should be starting to keep "scheduler note" (which might be a better thing to name it) as well as "survey name".  The survey name contents are likely butler-relevant, while the scheduler note contents are likely scheduler-relevant.

PR #293, opened 2023-03-13
replace Fields.db with text file

yoachim:
Replace the Fields.db data file and associated class with a text file that was (probably) the original source of the data.


ehneilsen (approved):
This all looks reasonable to me.

PR #294, opened 2023-03-14
U/yoachim/restore sched fix

yoachim:
Fix the `restore_scheduler` utility and add a unit test for it.

### DIFF
--- a/rubin_sim/scheduler/utils/utils.py
+++ b/rubin_sim/scheduler/utils/utils.py
@@ -149,7 +149,8 @@ def restore_scheduler(
     observatory : rubin_sim.scheduler.observatory.Model_observatory
         The observaotry object
     in_obs : np.array or str
-        Array of observation, or path to sqlite database to restore from
+        Array of observations (formated like rubin_sim.scheduler.empty_observation). If a string,
+        assumed to be a file and SchemaConverter is used to load it.
     filter_sched : rubin_sim.scheduler.scheduler object
         The filter scheduler. Note that we don't look up the official end of the previous night,
         so there is potential for the loaded filters to not match.
@@ -193,6 +194,7 @@ def restore_scheduler(
     observatory.observatory.current_dec_rad = obs["dec"]
     observatory.observatory.current_rot_sky_pos_rad = obs["rotSkyPos"]
     observatory.observatory.cumulative_azimuth_rad = obs["cummTelAz"]
+    observatory.observatory.current_filter = obs["filter"]
     observatory.observatory.mounted_filters = filters_needed
     # Note that we haven't updated last_az_rad, etc, but those values should be ignored.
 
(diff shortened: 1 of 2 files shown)


rubin_sim/scheduler/utils/utils.py
```diff
@@ -148,19 +148,24 @@ def restore_scheduler(
         Scheduler object.
     observatory : rubin_sim.scheduler.observatory.Model_observatory
         The observaotry object
-    filename : str
-        The output sqlite dayabase to use
+    in_obs : np.array or str
+        Array of observation, or path to sqlite database to restore from
```
rhiannonlynne, line 152: Could you clarify here if you need observations as came out of the scheduler, or observations as converted to the opsim database style using SchemaConverter? (or if either would work and the restore_scheduler will test which version it gets)

rubin_sim/scheduler/utils/utils.py
```diff
@@ -182,10 +187,11 @@ def restore_scheduler(
...
     )
+    observatory.obs_id_counter = obs["ID"] + 1
     observatory.observatory.parked = False
-    observatory.observatory.current_RA_rad = obs["RA"]
+    observatory.observatory.current_ra_rad = obs["RA"]
     observatory.observatory.current_dec_rad = obs["dec"]
-    observatory.observatory.current_rotSkyPos_rad = obs["rotSkyPos"]
+    observatory.observatory.current_rot_sky_pos_rad = obs["rotSkyPos"]
     observatory.observatory.cumulative_azimuth_rad = obs["cummTelAz"]
     observatory.observatory.mounted_filters = filters_needed
```
rhiannonlynne, line 196: I think you might also need to set observatory.observatory.current_filter to the one used in the last observation.

tests/scheduler/test_utils.py
```diff
@@ -3,10 +3,78 @@
...
+        mo, scheduler, observations = sim_runner(
+            mo,
+            scheduler,
+            survey_length=30.0,
+            verbose=False,
+            filename=None,
+            n_visit_limit=n_visit_limit,
+        )
+
+        # Won't be exact if we restart in the middle of a blob sequence, so need to restart on a new night
```
rhiannonlynne, line 35: Can you clarify this? Is it just because during a blob, the queue is generated all at once and not based on the actual conditions in the middle of the blob? (and does this mean we should mark "beginning of blobs"?)

rhiannonlynne (approved):
Have left some comments inline. One of these may require code changes,  another is a request, and the last is a question that may end up making potentially useful code changes (potentially in a different ticket).

tests/scheduler/test_utils.py
yoachim: Exactly, we can't reconstruct the queue (or any scheduled observations that were generated earlier in the night).

yoachim: It's usually easy to pick out the start of blob sequences since they are tagged with "a" or "b".

PR #295, opened 2023-03-15
Make survey_df more robust w.r.t. surveys without rewards

ehneilsen:
Uncovered this problem while experimenting with the new example scheduler. It works fine before a sim is done, but after this problem occurs.


yoachim (approved):
Yup, that seems in order. Should the `Survey` base class just set `self.reward=None` so you don't need to worry about checking if it exists?

ehneilsen:
The initializer of the base class already does set `self.reward=None`, but not all subclasses call the superclass initializer.

PR #297, opened 2023-03-21
U/yoachim/sched array restore

yoachim:
Introduce the new method `add_observations_array` throughout the code so surveys and features can be updated with arrays rather than requiring looping over observations individually. Results in ~50x speedup in restoring scheduler states.

Also expanded the example scheduler so it can be easily run from the command line to generate an example survey.

### DIFF
--- a/rubin_sim/scheduler/basis_functions/basis_functions.py
+++ b/rubin_sim/scheduler/basis_functions/basis_functions.py
@@ -82,7 +82,17 @@ class BaseBasisFunction(object):
         self.filtername = filtername
 
     def add_observations_array(self, observations_array, observations_hpid):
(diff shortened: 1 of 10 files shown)


rhiannonlynne:
In general more docstrings would be great. 
The motivation for having both the observation_array and observation_hpid isn't super obvious to me (given that the observation_hpid is the same and could be 'uniqued' back down if needed) .. but also, I imagine that it's faster to keep them both and once you have them just send them both around, so makes sense that way too. 

It would be nice if the update to the example survey / baseline survey was a separate PR for next time. (I know, I need to work on that too).

rhiannonlynne (approved):

PR #298, opened 2023-03-21
swap unit test to numpy allclose

yoachim:
Looks like there can be a slight difference in floating point values on different hardware. 

`(Pdb) norm_values["metric2"]["run9"]`
`-0.49167052452560656`
`(Pdb) ref_norm_values["metric2"]["run9"]`
`-0.49167052452560667`

Swapping to use numpy allclose so this passes.


ehneilsen (approved):
Looks fine to me.

PR #300, opened 2023-03-24
Simplify some slew time issues

rhiannonlynne:
Simply slew time calculation, modify slew time basis function so it returns Nans for unreachable areas, and stop replacement of array reward with a scalar (in cases with infinite values in reward) in BaseMarkovSurvey


yoachim (approved):
Yup, that all looks fine.

PR #301, opened 2023-03-28
U/yoachim/clouds update

yoachim:
Update the default cloud database to be the one generated in https://github.com/lsst/rubin_sim_notebooks/tree/main/clouds


ehneilsen (approved):
Looks good.

PR #302, opened 2023-03-29
Let 'baseline_run' be multiple runs

rhiannonlynne:
Letting the baseline_run be a list allows normalization of the summary metrics to the median value across that list. 
Specifying the list is better than just using all the values in the summary, as perhaps only some of the contents of the summary set should be used for the normalization, while others should just be normalized in the same way for plotting purposes. I.e. : send metrics for a set of weather/clouds + new test runs, but normalize only based on the weather/clouds runs while still normalizing the test runs on the same scale.

### DIFF
--- a/rubin_sim/maf/run_comparison/summary_plots.py
+++ b/rubin_sim/maf/run_comparison/summary_plots.py
@@ -86,18 +86,21 @@ def normalize_metric_summaries(
 
     # And now create a line just for "baseline" --
     # if baseline_run is >1, this is created from the median values per metric of those runs
-    if isinstance(baseline_run, list):
-        summary.loc["baseline"] = summary.loc[baseline_run].median(axis="rows")
-    else:
-        summary.loc["baseline"] = summary.loc[baseline_run]
+    # Make up a nonsense name for the reference, that is not currently in the summary dataframe
+    baseline_comparison = "bbb"
+    while baseline_comparison in summary.index:
+        baseline_comparison += "b"
 
-    baseline_run = "baseline"
+    if not isinstance(baseline_run, str):
+        summary.loc[baseline_comparison] = summary.loc[baseline_run].median(axis="rows")
+    else:
+        summary.loc[baseline_comparison] = summary.loc[baseline_run]
 
     if metric_sets is None:
         # If no invert/mag - just do simple normalization (1 + (x-0)/x0)
         norm_summary = 1 + (
-            summary.loc[:, :].sub(summary.loc[baseline_run, :], axis="columns")
-        ).div(summary.loc[baseline_run, :], axis="columns")
+            summary.loc[:, :].sub(summary.loc[baseline_comparison, :], axis="columns")
+        ).div(summary.loc[baseline_comparison, :], axis="columns")
     else:
         # Reindex metric set and remove duplicates or non-available metrics
         metric_names = [n for n in metric_sets.index.names if not n == "metric"]
@@ -116,36 +119,34 @@ def normalize_metric_summaries(
         )
 
         # Direct metrics are those that are neither inverted, nor compared as magnitudes
+        # direct = 1 + (value - norm) / norm == value / norm
         direct = ~np.logical_or(metric_sets["invert"], metric_sets["mag"])
         norm_summary.loc[:, direct] = summary.loc[:, direct]
 
+        # invert = 1 + (1/value - 1/norm) / (1/norm) == norm / value
         norm_summary.loc[:, metric_sets["invert"]] = (
             1.0 / summary.loc[:, metric_sets["invert"]]
         )
-        if baseline_run is not None:
-            norm_summary.loc[:, metric_sets["mag"]] = 1.0 + summary.loc[
-                :,
-                metric_sets["mag"],
-            ].subtract(summary.loc[baseline_run, metric_sets["mag"]], axis="columns")
-        else:
-            norm_vals = np.median(summary.loc[:, metric_sets["mag"]], axis=0)
-            norm_summary.loc[:, metric_sets["mag"]] = 1.0 + summary.loc[
-                :,
-                metric_sets["mag"],
-            ].subtract(norm_vals)
 
-        # Some metrics can be both inverted and magnitudes (eg rms mag values) - update
+        # mag = 1 + (1+value-norm - (1+norm-norm)) / (1+norm-norm) == 1 + (value - norm)
+        norm_summary.loc[:, metric_sets["mag"]] = 1.0 + summary.loc[
+            :,
+            metric_sets["mag"],
+        ].subtract(summary.loc[baseline_comparison, metric_sets["mag"]], axis="columns")
+
+        # Some metrics can be both inverted and magnitudes (eg rms mag values)
         both = np.logical_and(metric_sets["invert"], metric_sets["mag"])
+        # both = 1 + (1-(value-norm) - (1-(norm-norm))) / (1-(norm-norm)) == norm - value
         norm_summary.loc[:, both] = 1.0 - summary.loc[:, both].subtract(
-            summary.loc[baseline_run, both], axis="columns"
+            summary.loc[baseline_comparison, both], axis="columns"
         )
 
-        # Look a the fractional difference compared with the baseline
+        # Turn the values above into the fractional difference compared with the baseline
         norm_summary.loc[:, :] = 1 + (
             norm_summary.loc[:, :].sub(
-                norm_summary.loc[baseline_run, :], axis="columns"
+                norm_summary.loc[baseline_comparison, :], axis="columns"
             )
-        ).div(norm_summary.loc[baseline_run, :], axis="columns")
+        ).div(norm_summary.loc[baseline_comparison, :], axis="columns")
 
     # Set the index name
     norm_summary.columns.name = "metric"
--- a/rubin_sim/maf/run_comparison/summary_plots.py
+++ b/rubin_sim/maf/run_comparison/summary_plots.py
@@ -35,8 +35,9 @@ def normalize_metric_summaries(
 
     Parameters
     ----------
-    baseline_run : `str`
+    baseline_run : `str` or `list` of `str
         The name of the run that defines a normalized value of 1.
+        If a list is provided, the median value of each metric across that list is used as the reference.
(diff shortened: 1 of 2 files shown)


rubin_sim/maf/run_comparison/summary_plots.py
```diff
@@ -83,6 +84,15 @@ def normalize_metric_summaries(
     # Get rid of duplicate metrics and runs
     summary = summary.T.groupby("metric").first().T.groupby("run").first()
 
+    # And now create a line just for "baseline" --
+    # if baseline_run is >1, this is created from the median values per metric of those runs
+    if isinstance(baseline_run, list):
```
ehneilsen, line 89: Rather than checking whether `baseline_run` is a list, it would make the code more versatile to check if `summary.loc[baseline_run] `returns a single value or multiple values. Sometimes it will make more sense for a user to supply other sorts of collections, e.g. a `set`, `tuple`, or a `pandas.Series`, and supporting this would not make the code any more complex.

ehneilsen (approved):
See comments for a suggestion.

rubin_sim/maf/run_comparison/summary_plots.py
rhiannonlynne: Oh good idea ... although, this might be problematic as it is .. 
```  if len(summary.loc[baseline_run]) > 1:
/var/folders/tr/bjzqy0h17kx31fhndw6050d80000gn/T/ipykernel_16607/116705069.py:65: FutureWarning: Passing a set as an indexer is deprecated and will raise in a future version. Use a list instead.
```

rhiannonlynne: So: 
making the list a set instead gives me a future warning that it won't be supported in the future
making the list a tuple gives me key errors if the tuple is more than one element long (which I don't understand but may be a panda + me thing? it does seem like the same thing that works as a list, if the list is converted to a tuple, no longer works)
making the list a pd.Series using only the list does work

I'll make the adjustment because you're right that's a better and more flexible way to do it, but it's worth noting that set will work for now but fail later, and a tuple will not generally work as far as I can tell.  But at least the errors will be clearly pandas-related errors, which will hopefully lead people to realize they can't reference contents of a data frame in that way.

ehneilsen:
I think the important bit is not to introduce arbitrary and unnecessary constraints: if we can use it to index the DataFrame, why not accept it? In practice, I think the most likely is a pandas.Series, because that's often what you end up with when you're doing stuff with pandas do begin with.

One bit of advice I've seen on python in general is to notice if you are ever using isinstance, and if you are, think really carefully about whether you need to: it often implies that you are taking a "look before you leap" approach to a problem, while traditional idiomatic python prefers an "easier to ask forgiveness than permission" approach. Of course, sometimes LBYL really is the right was to go, but I find its generally worth thinking about explicitly on each case. In this case, the EAFP seems more general and without downsides.

From: Lynne Jones ***@***.***>
Date: Thursday, March 30, 2023 at 12:46 PM
Cc: "Eric H Neilsen Jr." ***@***.***>, Review requested ***@***.***>
Subject: Re: [lsst/rubin_sim] Let 'baseline_run' be multiple runs (PR #302)


@rhiannonlynne commented on this pull request.

________________________________

In rubin_sim/maf/run_comparison/summary_plots.py<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_lsst_rubin-5Fsim_pull_302-23discussion-5Fr1153591058&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=C1pT7tUhW3PFDWNL7ANIORYPZWXzstdq6bRpfbCu0ww&m=oIT-7KF7cfBCU94X6DoCLkSdiOrYXHFrex9e74a2mlwdij128XE-LtIK5DZc4X1r&s=eWpwd-djOYjRPaBAwlHCppqQD99ZRY49FaGAwtUTGRQ&e=>:

> @@ -83,6 +84,15 @@ def normalize_metric_summaries(

     # Get rid of duplicate metrics and runs

     summary = summary.T.groupby("metric").first().T.groupby("run").first()



+    # And now create a line just for "baseline" --

+    # if baseline_run is >1, this is created from the median values per metric of those runs

+    if isinstance(baseline_run, list):

Oh good idea ... although, this might be problematic as it is ..

/var/folders/tr/bjzqy0h17kx31fhndw6050d80000gn/T/ipykernel_16607/116705069.py:65: FutureWarning: Passing a set as an indexer is deprecated and will raise in a future version. Use a list instead.

rubin_sim/maf/run_comparison/summary_plots.py
rhiannonlynne: Also .. this may be me not knowing something about pandas, but if you provide only a string, then the summary.loc[baseline_run] returns a series, which has different kinds of length criteria than if you provided a list which then returns a data frame. 
Is there a benefit to checking if the returned value from "summaries.loc[baseline_run]" is a Series or a Dataframe, and then also checking the length of the Dataframe? Vs. just checking if baseline_run is a string (and if it's anything other than a string, assuming you will just go ahead and take the median .. if it's only a single element, taking the median would be the same as taking the row itself).

ehneilsen:
If it were me, I think I'd just check if it is a Series or DataFrame, and take the median if it's a DataFrame. If the user passes a list rather than a string, but that list is one element long, this means that it will take the median of a DataFrame one row long, but that's fine.

From: Lynne Jones ***@***.***>
Date: Thursday, March 30, 2023 at 1:23 PM
Cc: "Eric H Neilsen Jr." ***@***.***>, Review requested ***@***.***>
Subject: Re: [lsst/rubin_sim] Let 'baseline_run' be multiple runs (PR #302)


@rhiannonlynne commented on this pull request.

________________________________

In rubin_sim/maf/run_comparison/summary_plots.py<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_lsst_rubin-5Fsim_pull_302-23discussion-5Fr1153630373&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=C1pT7tUhW3PFDWNL7ANIORYPZWXzstdq6bRpfbCu0ww&m=0j62ByrEqiuVCfpKIU4cdXRgBtwaJicGsTvDFyhQsMyXNz2ZMumzLOlChErT1vTq&s=SK8f_pGFIrFdwsy9pIDr00G1hsYtkBFEKlZwdaEOsMc&e=>:

> @@ -83,6 +84,15 @@ def normalize_metric_summaries(

     # Get rid of duplicate metrics and runs

     summary = summary.T.groupby("metric").first().T.groupby("run").first()



+    # And now create a line just for "baseline" --

+    # if baseline_run is >1, this is created from the median values per metric of those runs

+    if isinstance(baseline_run, list):

Also .. this may be me not knowing something about pandas, but if you provide only a string, then the summary.loc[baseline_run] returns a series, which has different kinds of length criteria than if you provided a list which then returns a data frame.
Is there a benefit to checking if the returned value from "summaries.loc[baseline_run]" is a Series or a Dataframe, and then also checking the length of the Dataframe? Vs. just checking if baseline_run is a string (and if it's anything other than a string, assuming you will just go ahead and take the median .. if it's only a single element, taking the median would be the same as taking the row itself).

rhiannonlynne:
> I think the important bit is not to introduce arbitrary and unnecessary constraints: if we can use it to index the DataFrame, why not accept it? In practice, I think the most likely is a pandas.Series, because that's often what you end up with when you're doing stuff with pandas do begin with. One bit of advice I've seen on python in general is to notice if you are ever using isinstance, and if you are, think really carefully about whether you need to: it often implies that you are taking a "look before you leap" approach to a problem, while traditional idiomatic python prefers an "easier to ask forgiveness than permission" approach. Of course, sometimes LBYL really is the right was to go, but I find its generally worth thinking about explicitly on each case. In this case, the EAFP seems more general and without downsides. 

I may be missing something -- because pandas returns you either a data frame or a series, but they work quite differently, I think you still need to check which one you have (using isinstance) -- so you're either checking before to see which one you will get, or you're checking after to see which one you got. You can't even just cast the series into a data frame without checking, because the orientation/order will be incorrect (what would have been rows become columns in that case). 

I think checking which one you got back is reasonable though, and likely to catch potential problems easier if pandas changes what will return a Series instead of a DataFrame. I just don't think it's necessarily an applicable LBYL vs. EAFP here - you have to check in either case.

ehneilsen:
I was proposing to check after indexing, rather than before. That is, instead of:

if isinstance(i, str):
    v = df[i]
else:
    v = df[i].median(axis='rows')

do this:

v = df[i]
if not isinstance(v, pd.Series):
   v = df.median(axis='rows')

This way, you're directly testing on the behavior you want to react to (whether pandas returns a Series or DataFrame), rather than what you currently think might result in that behavior.

Yes, I'm still using "isinstance", but the point wasn't avoiding isinstance, only improving generality.

	-Eric

From: Lynne Jones ***@***.***>
Date: Thursday, March 30, 2023 at 2:13 PM
Cc: "Eric H Neilsen Jr." ***@***.***>, Review requested ***@***.***>
Subject: Re: [lsst/rubin_sim] Let 'baseline_run' be multiple runs (PR #302)

I think the important bit is not to introduce arbitrary and unnecessary constraints: if we can use it to index the DataFrame, why not accept it? In practice, I think the most likely is a pandas.Series, because that's often what you end up with when you're doing stuff with pandas do begin with. One bit of advice I've seen on python in general is to notice if you are ever using isinstance, and if you are, think really carefully about whether you need to: it often implies that you are taking a "look before you leap" approach to a problem, while traditional idiomatic python prefers an "easier to ask forgiveness than permission" approach. Of course, sometimes LBYL really is the right was to go, but I find its generally worth thinking about explicitly on each case. In this case, the EAFP seems more general and without downsides.
I may be missing something -- because pandas returns you either a data frame or a series, but they work quite differently, I think you still need to check which one you have (using isinstance) -- so you're either checking before to see which one you will get, or you're checking after to see which one you got. You can't even just cast the series into a data frame without checking, because the orientation/order will be incorrect (what would have been rows become columns in that case).
I think checking which one you got back is reasonable though, and likely to catch potential problems easier if pandas changes what will return a Series instead of a DataFrame. I just don't think it's necessarily an applicable LBYL vs. EAFP here - you have to check in either case.

PR #304, opened 2023-04-04
Add option to record rewards during simulation

ehneilsen:
I added options to record basis functions and rewards during a simulation, so that they can then be plotted in dashboards.
What is here may eventually need to be supplemented so that there is a direct way to figure out where long gap observations came from.

I really only intend for this to be used in simulations of a few nights or less, although I suppose we could try with a full simulation to see if it is practical in practice.

Note that the "accum" options was added so that it can be turned off, because it increases the time of simulation by more than a factor of 6, and is generally not necessary. If we really wanted it, some optimization could be done by caching basis function values, but at this point I don't think it worth the added work and code complexity.

### DIFF
--- a/rubin_sim/scheduler/schedulers/core_scheduler.py
+++ b/rubin_sim/scheduler/schedulers/core_scheduler.py
@@ -2,11 +2,13 @@ from __future__ import absolute_import
 import reprlib
 from collections import OrderedDict
 from io import StringIO
+import time
 from copy import deepcopy
 from builtins import object
 import numpy as np
 import healpy as hp
 import pandas as pd
+from astropy.time import Time
 from rubin_sim.utils import _hpid2_ra_dec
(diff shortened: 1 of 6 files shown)


yoachim (approved):
That all seems good. And it looks like the `accum` doesn't kick in by default in sim_runner, so I shouldn't see an unexpected run time jump.

ehneilsen:
Any idea why the test is failing on ubuntu (but succeeding on macos)? It doesn't look related to anything in this pull request.

ehneilsen:
Also, _none_ of it should kick in unless explicitly requested when starting sim_runner. If it does, that's a bug.

yoachim:
That looks like a weird transient thing in the moving objects code. I say just merge it, if it fails after that I'll poke around on it.

PR #305, opened 2023-04-05
U/yoachim/long on add

yoachim:
Switch around how long gaps does it's accounting. Now should work if a scheduler is reconstructed in the middle of the night.


ehneilsen (approved):
Looks okay, as far as I can tell.

PR #306, opened 2023-04-05
U/yoachim/more kne

yoachim:


rhiannonlynne:
What was the increase in time required to run the metric? (for documentation purposes).

PR #307, opened 2023-04-06
Add unique readable names for surveys and basis functions to the data…

ehneilsen:
…frame of basis function values recorded in rubin_sim

This will make it easier for use by schedview, or anything else that wants to look at the same data.

The code that actually creates the labels here may need to be updated at the conclusion of [RFC-913](https://jira.lsstcorp.org/browse/RFC-913),  but given the interface here, schedview shouldn't care how its implemented, and whatever changes will be needed should be easy.


yoachim (approved):
Seems straight forward

PR #308, opened 2023-04-07
remove extra alt and az trimming in BlobSurvey

yoachim:
Remove the altitude, azimuth, and radius cuts from `BlobSurvey`. The altitude cut is already done by a basis function in our usual setup. The radius and azimuth cuts are nominally unnecessary (a test shows we lose 0.1% of visits when they are removed). Put a `**kwargs` in there so this is backwards compatible, old kwargs just ignored if they are specified.


ehneilsen (approved):
Looks good

PR #309, opened 2023-04-12
U/yoachim/per year coadd

yoachim:
Add requested coadds computed per year to batches. 
Fix bug in log gap survey that was making gaps too short.


ehneilsen (approved):
Looks fine.

PR #310, opened 2023-04-14
U/yoachim/tsp failing

yoachim:
Put a radius limit on building blobs back in, but only apply it after the reward function is computed. This prevents an occasional failure with the TSP solver.


ehneilsen (approved):
Looks fine

PR #311, opened 2023-04-14
Mask on alt before taking max. and area when recording basis function…

ehneilsen:
… values

### DIFF
--- a/rubin_sim/scheduler/surveys/base_survey.py
+++ b/rubin_sim/scheduler/surveys/base_survey.py
@@ -8,6 +8,9 @@ from rubin_sim.scheduler.utils import (
     HpInComcamFov,
     comcam_tessellate,
 )
+from rubin_sim.scheduler.basis_functions.mask_basis_funcs import (
+    ZenithShadowMaskBasisFunction,
+)
 from rubin_sim.site_models import _read_fields
 import healpy as hp
 from rubin_sim.scheduler.thomson import xyz2thetaphi, thetaphi2xyz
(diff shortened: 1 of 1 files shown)


ehneilsen:
Using all the masked areas instead of just ZenithShadowMaskBasisFunction would defeat some of the purpose, which is to be able to figure which basis functions are causing a given survey not to be selected. One of the statistics kept is the area on the sky left unmasked after each basis function. But, if we applied all masks on top of each other to all basis functions, this statistic would be the same for all, and it wouldn't really tell us anything.

One possible alternative is not to use ZenithShadowMaskBasisFunction at all, but rather just apply a zenith distance cut directly, independent of any basis functions in the survey. This would make it more robust to things like ZenithShadowMaskBasisFunction being renamed, but in effect create some code duplication between this function and code that is already in the implementation of ZenithShadowMask. Do you think this would be better?


From: Peter Yoachim ***@***.***>
Date: Monday, April 17, 2023 at 11:18 AM
Subject: Re: [lsst/rubin_sim] Mask on alt before taking max. and area when recording basis function… (PR #311)


@yoachim approved this pull request.

Would it make sense to pull all the masked areas from all the basis functions? I'm just worried if someone goes and renames ZenithShadowMaskBasisFunction or uses something similar but named different than this will then mysteriously fail. Or it could throw a warning if it doesn't find a zenith mask I guess.

ehneilsen:
I just replaced the use of an existing zenith mask basis function with just making my own low alt mask generated at the time I need it.

yoachim (approved):

PR #316, opened 2023-04-28
add sun alt limit to scheduled observations

yoachim:
Update scheduled observations so a sun altitude limit can be set. Also decrease the default number of observations `ScriptedSurvey` returns to prevent DDF observations running into twilight time.


ehneilsen (approved):
Looks good.
Was there not a sun alt limit already?

yoachim:
When initially scheduled there is a sun limit, but if downtime (or another DDF) bumps it out of the initial ideal time slot it'll keep trying and can then run into twilight.

PR #319, opened 2023-05-09
patch long gaps so they can only execute once per night

yoachim:
Add a basis function so long gap blobs will only execute once in a night. Also make sure long gap observations don't bump into twilight time.

### DIFF
--- a/rubin_sim/scheduler/basis_functions/feasibility_funcs.py
+++ b/rubin_sim/scheduler/basis_functions/feasibility_funcs.py
@@ -7,6 +7,7 @@ from rubin_sim.scheduler.utils import IntRounded
 
 __all__ = [
     "FilterLoadedBasisFunction",
+    "OnceInNightBasisFunction",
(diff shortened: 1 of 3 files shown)


rubin_sim/scheduler/basis_functions/feasibility_funcs.py
```diff
@@ -51,6 +52,26 @@ def check_feasibility(self, conditions):
         return result
 
 
+class OnceInNightBasisFunction(BaseBasisFunction):
```
rhiannonlynne, line 55: I suspect tagging things as "feasibility function that's a boolean" vs. "basis function with a value" and then maybe also as spatial or single value is going to be handy sooner rather than later, just for easier readability.

rhiannonlynne:
Looks fine - do you have an example of where this was broken and that now it's working?

rhiannonlynne (approved):

PR #323, opened 2023-05-15
let all area be masked

yoachim:
Fixes problem when running an "area required" survey ends up with all area masked.


ehneilsen (commented):
Looks good.

ehneilsen (approved):

PR #325, opened 2023-05-25
Tickets/opsim 1023

yoachim:
Refactor of `moving_objects`.
* Updated output to be a npz file written all at once rather than continuous writes to txt
* Option for using pre-computed orbital positions which uses dynamic detection tolerance
* general compacting of moving_objects.utils


rubin_sim/moving_objects/make_lsst_obs.py
```diff
@@ -209,34 +216,69 @@ def make_lsst_obs():
...
         args.simulation_db,
         colmap,
         constraint=args.sql_constraint,
-        footprint=args.footprint,
         dbcols=None,
     )
-    # Generate ephemerides.
-    mo.run_obs(
+
+    d_obs = mo.DirectObs(
```
rhiannonlynne, line 239: Since this is hard-coded here, it makes sense to remove the option to use LinearObs in the arguments. 
I mentioned earlier removing LinearObs as a class, but I think you could leave it -- but it should match the API for DirectObs, so would need an update so that it has the same API. (the point being that someone else could make a different method for calculating the observations .. such as using SpicePy or whatever .. and then just drop it in in place of DirectObs, only rewriting the "run" method).

rubin_sim/moving_objects/make_lsst_obs.py
```diff
@@ -42,6 +43,12 @@ def setup_args(parser=None):
         "See https://github.com/oorb/oorb/tree/master/python#defining-orbits for "
         "additional documentation on the orbit file format. Default None.",
     )
+    parser.add_argument(
+        "--positions_file",
+        type=str,
+        default=None,
+        help="file with pre-computed positions for the objects in the orbit file",
+    )
```
rhiannonlynne, line 51: Could you please match the "help" string with the others in this file? 
"File with pre-computed 'rough' ephemerides for the objects in the orbit file. Default None, which will trigger computation of these rough ephemerides on-the-fly."

rubin_sim/moving_objects/make_lsst_obs.py
```diff
@@ -82,7 +89,7 @@ def setup_args(parser=None):
         type=float,
         default=1.75,
         help="If using a circular footprint, this is the radius of the FOV (in degrees). "
-        "Default 1.75 deg.",
+        "Default 2.08 deg.",
```
rhiannonlynne, line 92: The default value for r_fov should be 1.75 degrees (sorry if this is the result of miscommunication on slack). 
The use of r_fov should be for defining the circular footprint, not for defining the matching tolerance for rough ephemerides. Can you please update the docstring to match the default value?

rhiannonlynne: Is there an "outfile name" at some point in the arguments here? Probably needs an update to match the .npz filetype now?

rubin_sim/moving_objects/make_lsst_obs.py
```diff
@@ -209,34 +216,69 @@ def make_lsst_obs():
         logging.basicConfig(level=logging.INFO)
 
     # Read orbits.
-    orbits = mo.read_orbits(args.orbit_file)
+    orbits = mo.Orbits()
+    orbits.read_orbits(args.orbit_file)
 
+    if args.positions_file is not None:
+        position_data = np.load(args.positions_file)
+        object_positions = position_data["positions"].copy()
```
rhiannonlynne, line 224: For my own education - why does this need to be a copy?

rubin_sim/moving_objects/pre_generate.py
```diff
@@ -0,0 +1,39 @@
...
+    types = [float] * 2
+    dt = list(zip(names, types))
+
+    for filename in orbit_files:
+        print("working on %s" % filename)
+        orbits = Orbits()
+        orbits.read_orbits(filename)
+        # Array to hold results
+        results = np.zeros((len(orbits.orbits), np.size(mjds)), dt)
+        do = DirectObs()
```
rhiannonlynne, line 27: May I suggest using PyOrbEphemerides directly to generate these ephemerides? That way you don't have to do any FOV checking or calculate colors or trailing losses. There is an example here - https://github.com/lsst/rubin_sim_notebooks/blob/main/moving_objects/moving_objects_demo_part_1.ipynb

rubin_sim/maf/generate_ss.py
```diff
@@ -82,9 +94,12 @@ def generate_ss_commands(
             for run, filename in zip(runs, dbfiles):
                 objtype = objtypes[pop]
 
-                s1 = f"make_lsst_obs --simulation_db {filename} --orbit_file {orbit_files[pop]}"
+                s1 = (
+                    f"make_lsst_obs --simulation_db {filename} --orbit_file {orbit_files[pop]}"
+                    f" --positions_file {position_files[pop]}"
+                )
```
rhiannonlynne, line 100: These are relatively minor changes, can you please add them to the "split" processing option as well? 
(I am curious how split processing might work if we have lots of cores and can more interactively split the orbit files into multiple subsets depending on how many workers are available -- i.e. not just 10, but maybe 100 subsets -- and reduce the processing time per simulation. In the future we're likely to have many fewer simulations, and need to turn them around faster).

rhiannonlynne (changes requested):
Some relatively minor comments.

rubin_sim/moving_objects/direct_obs.py
```diff
@@ -121,27 +130,19 @@ def __init__(
...
-        self.outfile.write(
-            "# ephMode %s prelimEphMode %s\n" % (self.eph_mode, self.prelim_eph_mode)
-        )
-        self.outfile.write(
-            "# rough tolerance for preliminary match %f\n" % self.rough_tol
-        )
-        self.outfile.write("# time step for preliminary match %f\n" % self.tstep)
-
-    def run(self, orbits, obs_data):
+    def run(self, orbits, obs_data, object_positions=None, object_mjds=None):
```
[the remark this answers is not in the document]
yoachim: OK, info tracked and saved. All the old camera and footprint options are still there, but the interpolation option for positions is gone.

rubin_sim/moving_objects/direct_obs.py
```diff
@@ -153,52 +154,165 @@ def run(self, orbits, obs_data):
             The orbits to generate ephemerides for.
         obs_data : `np.ndarray`
             The simulated pointing history data.
+        object_positions : `np.ndarray`
+            Pre-computed RA,dec positions for each object in orbits (degrees)
```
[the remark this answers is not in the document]
yoachim: Not really. Just dense enough that you don't have to worry about the object completing a retrograde loop so it looks like it didn't travel far in the time step when it actually went many degrees.

rubin_sim/moving_objects/direct_obs.py
```diff
@@ -153,52 +154,165 @@ def run(self, orbits, obs_data):
...
+            "trueAnomaly",
+            "velocity",
+            "dmag_color",
+            "dmag_trail",
+            "dmag_detect",
+        ]
+        types = [int, int, "<U40"] + [float] * (len(names) - 3)
+
+        # XXX--for now, copy over some observation info. In the future,
+        # just index it to the observations and only transfer observationId
```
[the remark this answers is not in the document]
yoachim: In theory, as long as you have the observation ID you could just re-load the observations and say things like `seeing = observations_array[seeing_col][object_array[observationId]]`. You get to drop some columns from the output, but then you are forced to keep track of the database file, so kind of a tradeoff. Probably only worth it if we do more downstream refactoring.

rubin_sim/moving_objects/make_lsst_obs.py
yoachim: The `.copy()` loads it into memory, otherwise it'll read it off disk every time it gets referenced. And maybe it can't be accessed after doing a `.close()` if there's no `.copy()`

rubin_sim/moving_objects/pre_generate.py
yoachim: It's calling `generate_ephemerides` not `run`, so there is no extra checking or trailing losses calculated. I like using `DirectObs` because then I know these positions were created with the default kwargs like `obs_time_scale`, so as long as I'm using defaults later it will all match up.

rubin_sim/maf/generate_ss.py
yoachim: Sure. I think we can simplify the scatter/gather process for running things in parallel, but that can be a different ticket.

rubin_sim/moving_objects/pre_generate.py
rhiannonlynne: Reasonable. It might make sense to add some information like the timescale into the saved files, or at least some information in the README associated with those files about some of these details.

rubin_sim/moving_objects/make_lsst_obs.py
rhiannonlynne: LinearObs was removed.

rhiannonlynne (approved):

PR #326, opened 2023-05-30
Ticket/opsim 1029 Set survey name and add target column to observations

yoachim:
Add a method so that `survey_name` is always set
Add `target` as an output column


rhiannonlynne:
I'm curious how `target` gets set, in order to send it to the observation output? 
Can we add something as example, such as automatically setting target to be "DD <fieldname>" for the DDFs?

rhiannonlynne:
Oh and another question because this wasn't obvious to me before .. 
using the survey_name as the "note" (which is what I think is happening) .. does this mean that the survey_name will automatically be propagated to features which need to know that in order to properly ignore values? 
And does this work when the blob survey_name is "Blob r_i" (and not "blob, ri, a" or "blob, ri, b" etc)?

rhiannonlynne (approved):

yoachim:
> Oh and another question because this wasn't obvious to me before .. using the survey_name as the "note" (which is what I think is happening)

The survey_name attribute does not automatically propagate to the "note" field. The base survey class now has a `_generate_survey_name` method, so any subclass can set up exactly how it wants to construct it's name. The observation `note` and `target` fields can be set in the `generate_observations` method. Some of the surveys are just passing on `survey_name`, others like the blobs construct more complicated notes.

rhiannonlynne:
That's great. The separation of survey_name (for things like schedview) and note is good.

PR #327, opened 2023-06-01
OPSIM-1033 : Add GeneralUniformityMetric

rhiannonlynne:
A generalization of the UniformityMetric, beyond MJD. 
This was motivated because I saw someone calculating uniformity for a different derivable data column (I think it was phase for a periodic star), and they wrote their own version of the UniformityMetric.

### DIFF
--- a/rubin_sim/maf/metrics/cadence_metrics.py
+++ b/rubin_sim/maf/metrics/cadence_metrics.py
@@ -4,6 +4,7 @@ from .base_metric import BaseMetric
 __all__ = [
     "TemplateExistsMetric",
     "UniformityMetric",
+    "GeneralUniformityMetric",
     "RapidRevisitUniformityMetric",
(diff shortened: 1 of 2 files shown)


yoachim (approved):
Looks good.

It's tempting to swap this and the other similar metrics to return `np.abs(result-1)` so that 1 is perfectly uniform and 0 is a delta function. I never remember to invert it at least.

rhiannonlynne:
Ah, that's a good point. "1" == "good" and uniform, 0 == "bad" and nonuniform. 
Would it be too confusing to swap just this new one and not the old one? 
I am not sure if anyone else is using these results anywhere, but I would suspect not.

yoachim:
DESC was looking at the Kuiper metric before and I think there was confusion since it was defined where smaller was better. That might be the only other one that needs flipping.

PR #328, opened 2023-06-02
Update basis functions for v3.2

yoachim:
* update the NObsPerYear basis function so it can turn off after first year
* add a solar elongation mask


ehneilsen (approved):
Looks fine.

rubin_sim/scheduler/basis_functions/mask_basis_funcs.py
```diff
@@ -19,6 +20,29 @@
 ]
 
 
+class SolarElongMaskBasisFunction(BaseBasisFunction):
+    """Mask regions beyond some solar elongation
```
rhiannonlynne, line 24: Could you please make it clear in the docstring if "beyond some solar elongation" means "closer to the sun" (smaller values) or "farther from the sun" (larger values)? 
I would have thought that it might be reasonable to add both upper and lower limits to the solar elongation mask? 
(don't have to set them both)

PR #329, opened 2023-06-16
remove read for moving objects test

yoachim:
Writing and reading cheby coefficients was having a mysterious OS-dependent failure. Probably just a Jenkins issue. Unit test updated to skip the writing/reading step be passing coefficients directly.

### DIFF
--- a/rubin_sim/moving_objects/cheby_fits.py
+++ b/rubin_sim/moving_objects/cheby_fits.py
@@ -575,6 +575,11 @@ class ChebyFits(object):
         append : bool, optional
             Flag to append (or overwrite) the output files.
         """
+
+        warnings.warn(
+            "Writing cheby fit values may have cross-platform issues. Consider passing values directly"
+        )
+
         if append:
             open_mode = "aw"
         else:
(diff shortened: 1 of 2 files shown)


rhiannonlynne:
Was the problem the 'read' instead of 'set' or was it actually the fact that there was Nans / not-fit ephemerides in the coefficient file?

yoachim:
Looks like both. One test passed by swapping read to set, the other was still failing to fit and throwing NaNs.

rhiannonlynne:
I'm somewhat confused about why the read instead of set is causing an error then .. there is a separate unit test that checks if the set vs. read gives "approximately the same" coefficients in cheby_values. 
It's *possible* that something has changed such that now the tolerance allowed in that process results in different values in the resulting calculated ephemerides, but it seems strange that it's only on ubuntu and only now. 

Anyway, I guess maybe the thing to do for now is to make a short warning in the "write" method of cheby_fits along the lines of "Warning: it is possible that cross-platform differences in writing with cheby_fits and then reading/using the coefficients in cheby_values (compared to just passing the coefficients from cheby_fits into cheby_values) can cause noticeable differences in the resulting ephemerides. Further checking for a particular use-case is suggested."

rhiannonlynne (approved):

PR #330, opened 2023-06-16
make it possible to replace seeing, clouds, downtime in ModelObservatory

yoachim:
Make it possible to add in arbitrary cloud, seeing, and downtime to the model observatory.


ehneilsen (approved):
This looks like it will do what I need it to.

I think the naming, separating seeing_data from the seeing model, might be a little confusing without context. For example, it looks like I can pass a function that simulates clouds or seeing with a stochastic model instead of a data set as "cloud_data" or "seeing_data" (and I expect to actually write such simulators), but intuitively this seems like more of a model to me than data. But really, this is a minor quibble.

PR #332, opened 2023-06-22
Tickets/opsim 1040 update example scheduler

yoachim:
Update the example scheduler to be the same as the v3.2 baseline.

### DIFF
--- a/rubin_sim/scheduler/example/example_scheduler.py
+++ b/rubin_sim/scheduler/example/example_scheduler.py
@@ -1,5 +1,7 @@
 import numpy as np
+import matplotlib.pylab as plt
 import healpy as hp
+from rubin_sim.scheduler.model_observatory import ModelObservatory
 from rubin_sim.scheduler.schedulers import CoreScheduler, FilterSchedUzy
 from rubin_sim.scheduler.utils import (
     EuclidOverlapFootprint,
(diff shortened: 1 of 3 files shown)


rubin_sim/scheduler/example/example_scheduler.py
```diff
@@ -14,24 +16,191 @@
...
+    season=300.0,
+    season_start_hour=-4.0,
+    season_end_hour=2.0,
+    moon_distance=30.0,
+):
+    """Generate the standard basis functions that are shared by blob surveys
+
+    Parameters
+    ----------
+    XXX
```
ehneilsen, line 61: Add the documentation here...

rubin_sim/scheduler/example/example_scheduler.py
```diff
@@ -14,24 +16,191 @@
...
+
+
+    Returns
+    -------
+    list of tuples pairs (basis function, weight) that is
```
ehneilsen, line 66: If we're following DM's lead, we should follow the way that define types, [here](https://developer.lsst.io/python/numpydoc.html#py-docstring-returns) and [here](https://developer.lsst.io/python/numpydoc.html#py-docstring-parameter-types)

rubin_sim/scheduler/example/example_scheduler.py
```diff
@@ -1169,18 +1030,27 @@ def generate_twi_blobs(
...
-    Paramters
-    ---------
-    dist_to_eclip : float (40)
-        How far to extend from the ecliptic (degrees). default 40.
-    dec_max : float (30)
-        The maximum declination to extend to (degrees). default 30.
-    mask : np.array (None)
-        A HEALpix map that the result is multiplied by (default None)
-    """
+    """Generate a target_map for the area around the ecliptic"""
```
ehneilsen, line 1053: Why drop the docstring for the parameters? If it's private enough that the docstring isn't needed, we should rename the function with an underscore prefix (`_ecliptic_target`).

rubin_sim/scheduler/example/example_scheduler.py
```diff
@@ -1503,27 +1382,28 @@ def example_scheduler(
...
     ddfs = ddf_surveys(
         detailers=details,
-        season_unobs_frac=ddf_season_unobs_frac,
+        season_unobs_frac=ddf_season_frac,
         euclid_detailers=euclid_detailers,
     )
 
-    greedy = gen_greedy_surveys(nside, nexp=nexp, footprints=footprints)
-    neo = generate_twilight_neo(
+    greedy = gen_GreedySurveys(nside, nexp=nexp, footprints=footprints)
```
ehneilsen, line 1391: Why the name change to camel case?

ehneilsen (changes requested):
A little docstring work is needed.

ehneilsen (approved):

PR #333, opened 2023-06-27
OPSIM-1047 automatic backup URL for data

rhiannonlynne:
Failover to rubin-sim-data @ epyc. 
Add progress bar. 
Write file as it's read, instead of waiting to the end.
Only implemented on rs_download_data (but should be extended to rs_download_sky .. not on epyc yet).


ehneilsen (approved):
I don't see any problems.

PR #337, opened 2023-07-11
example scheduler to any nside

yoachim:
Make sure example scheduler can run with arbitrary nside and update unit test so it stays that way.


ehneilsen (approved):
Looks good, thanks!

PR #351, opened 2023-08-16
U/yoachim/radar redacted

yoachim:
Fixed a hodgepodge of bugs. Added a new science batch so that it's easy to run everything but the super time intensive ones.

### DIFF
--- a/rubin_sim/maf/batches/radar_limited.py
+++ b/rubin_sim/maf/batches/radar_limited.py
@@ -1,0 +1,1079 @@
+__all__ = ("radar_limited",)
+
+import astropy.units as u
+import healpy as hp
+import numpy as np
+
+import rubin_sim.maf as maf
+import rubin_sim.maf.maps as maps
+import rubin_sim.maf.metric_bundles as mb
+import rubin_sim.maf.metrics as metrics
+import rubin_sim.maf.plots as plots
+import rubin_sim.maf.slicers as slicers
+
+from .common import extended_summary, filter_list, lightcurve_summary, standard_summary
+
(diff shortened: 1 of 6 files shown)


rubin_sim/maf/batches/radar_limited.py
```diff
@@ -0,0 +1,1079 @@
...
+        allfilterinfo_label,
+    ) = filter_list(all=True)
+
+    standardStats = standard_summary(withCount=False)
+
+    # This is the default slicer for most purposes in this batch. Note that the cache is off -
+    # if the metric requires a dust map, this is not the right slicer to use.
+    healpixslicer = slicers.HealpixSlicer(nside=nside, use_cache=True)
+    subsetPlots = [plots.HealpixSkyMap(), plots.HealpixHistogram()]
+
```
ehneilsen, line 55: Stylistically, the code could be clearer and more reusable if the different sections (marked off with `##########` blocks) were extracted into separate functions, and the `radar_limited` function just call each in turn.

ehneilsen (approved):
Just a couple of minor, completely optional, style suggestions.

rubin_sim/maf/batches/radar_limited.py
yoachim: Normally I agree. In this case the code is generating output for MAF, so I really like having all the code that generates numbers and plots for a web page in one place. Otherwise when someone says "change this colorbar to log", or "was this configured for 3 or 4 points before the peak of the light curve" it becomes a nightmare to trace things back to the source.

PR #352, opened 2023-08-23
Tickets/opsim 1065

yoachim:
Fix issue in OneDSlicer where right edge was open rather then closed.


rhiannonlynne:
Looks good. The unit tests were failing due to black for the updated files, so I just ran that and pushed. 
I did also add an explicit test for the one_d slicer based on an integer 'night' column, just to be sure that we can actually define night based on 0-1, 1-2, etc. as well as 0.5-1.5, 1.5-2.5, etc.

rhiannonlynne (approved):

PR #353, opened 2023-08-25
make sure conditions object resets if mjd is changed

yoachim:
Also get rid of a few number deprecation warnings and isort things.


ehneilsen (approved):
Looks fine. This way, at least, if someone resets the mjd and then tries using it without updating everything they need to, there should be an error of one sort on another.
There may also be value in adding a warning if the mjd setter gets called.

PR #355, opened 2023-09-07
Update throughputs and m5 values

rhiannonlynne:
Update to v1.9 of the throughputs (3ag mirrors, updated to measured mirror reflectivity, broadband coatings for the lenses, and filter throughputs).


yoachim (approved):
wow, u band does change a lot

PR #356, opened 2023-09-07
OPSIM-1069: fix m5 diff basis function

yoachim:
Fix bug where M5Diff was using the dark sky brightness rather than the best possible 5-sigma limiting depth. Also update model observatory to create fresh Conditions objects to prevent excessive warnings.

### DIFF
--- a/rubin_sim/scheduler/model_observatory/model_observatory.py
+++ b/rubin_sim/scheduler/model_observatory/model_observatory.py
@@ -1,5 +1,6 @@
 __all__ = ("ModelObservatory", "NoClouds", "NominalSeeing")
 
+
 import healpy as hp
 import numpy as np
 from astropy.coordinates import EarthLocation
@@ -215,11 +216,12 @@ class ModelObservatory:
         sun_moon_info = self.almanac.get_sun_moon_positions(self.mjd)
         season_offset = create_season_offset(self.nside, sun_moon_info["sun_RA"])
         self.sun_ra_start = sun_moon_info["sun_RA"] + 0
+        self.season_offset = season_offset
         # Conditions object to update and return on request
         self.conditions = Conditions(
             nside=self.nside,
-            mjd_start=mjd_start,
-            season_offset=season_offset,
+            mjd_start=self.mjd_start,
+            season_offset=self.season_offset,
             sun_ra_start=self.sun_ra_start,
         )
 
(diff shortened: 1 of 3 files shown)


rhiannonlynne:
Hey, so maybe a stupid question - but why *aren't* we just comparing the current skybrightness maps against the dark skybrightness values? 
This would eliminate the effect of changing seeing resulting in a different overall value for the basis function, and it would still have the desired effect of guiding observations towards the parts of the sky which are darkest and into the filters which would work best at this point in the lunar cycle / position on the sky. 
And would require fewer calculations.

rubin_sim/scheduler/model_observatory/model_observatory.py
```diff
@@ -215,11 +216,13 @@ def __init__(
         sun_moon_info = self.almanac.get_sun_moon_positions(self.mjd)
         season_offset = create_season_offset(self.nside, sun_moon_info["sun_RA"])
         self.sun_ra_start = sun_moon_info["sun_RA"] + 0
+        self.mjd_start = mjd_start
```
ehneilsen, line 219: Isn't this already done more reliably in line 121?

ehneilsen (approved):
Double check line 219 (and/or make sure it behaves properly when mjd_start is None), but otherwise, looks fine.

yoachim:
> Hey, so maybe a stupid question - but why _aren't_ we just comparing the current skybrightness maps against the dark skybrightness values? This would eliminate the effect of changing seeing resulting in a different overall value for the basis function, and it would still have the desired effect of guiding observations towards the parts of the sky which are darkest and into the filters which would work best at this point in the lunar cycle / position on the sky. And would require fewer calculations.

If we only used the dark sky map we would lose the airmass extinction term. By using the full 5-sigma limiting depth, the scheduler "knows" that sticking close to the meridian is much more important in u than in z. 

I think we want the basis function to change value with different seeing conditions. That's a good thing. It means we have to set a fiducial seeing value, which is kind of an arbitrary magic number. But getting that value wrong is only a few tenths of a mag difference, while airmass and skybrightness terms vary by several mags.

rubin_sim/scheduler/model_observatory/model_observatory.py
yoachim: Indeed!

PR #363, opened 2023-09-22
move dust maps data to scheduler

yoachim:
Move copies of the HEALpix dust extinction maps to the scheduler data so the scheduler does not rely on `maps` data anymore
update sim_baseline to v3.3
clean up throughputs data (it had hidden git repo directories, etc)


ehneilsen (approved):
Looks good to me.

PR #367, opened 2023-10-03
Tickets/opsim 1081

yoachim:
Extend almanac so we can run up to 20 years.


rhiannonlynne:
The changes are minor and a good idea. 
In order to work on getting our code working with ruff and the documentation, could you please go through and update at least these files that were touched in the PR? ruff has some small errors. 
There are no docstrings .. which given that they're scripts, I'm not entirely sure where they go, but perhaps you could find a place to put a general description of the purpose or use of the script?

rhiannonlynne (approved):

PR #369, opened 2023-10-05
Remove six

rhiannonlynne:
There was a remaining six import I missed. 
I'm actually wondering if this entire class should be removed, but it might have to wait until we're ready to remove ObservationMetaData entirely.

### DIFF
--- a/rubin_sim/utils/spatial_bounds.py
+++ b/rubin_sim/utils/spatial_bounds.py
@@ -30,7 +30,7 @@ class SpatialBoundsMetaClass(type):
         return super(SpatialBoundsMetaClass, cls).__init__(name, bases, dct)
 
 
-class SpatialBounds(metaclass=SpatialBoundsMetaClass:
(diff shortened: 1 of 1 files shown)


rubin_sim/utils/spatial_bounds.py
```diff
@@ -31,7 +30,7 @@ def __init__(cls, name, bases, dct):
         return super(SpatialBoundsMetaClass, cls).__init__(name, bases, dct)
 
 
-class SpatialBounds(with_metaclass(SpatialBoundsMetaClass, object)):
+class SpatialBounds(metaclass=SpatialBoundsMetaClass:
```
ehneilsen, line 33: Can this really run at all? It looks like the parenthesis are unbounded.

ehneilsen (approved):

rubin_sim/utils/spatial_bounds.py
rhiannonlynne: Thank you .. that's what you get for editing quickly in emacs in the terminal.

PR #372, opened 2023-10-14
OPSIM-1086: remove obsolete code and simplify upkeep of example_scheduler

rhiannonlynne:
This PR was started to simplify some of the code in the footprints file (mostly, removing functions which were unused at this time and replaced by SkyAreaGenerators). Removing code which has names like `wfd_healpixels` which could tempt users to try to define the WFD by these functions -- but that we don't keep up to date with current definitions -- seems like good preventative maintenance. 

And then I removed some extraneous code from the example_scheduler file, and wanted to be able to use the arg parser to set standard arguments for the baseline survey while using a notebook .. and maybe to keep the example_scheduler up to date with the evolving baseline more easily. So there are changes in this file as well.


yoachim (approved):
It'd be nice if the big docstring/comment changes were separate from the code changes.

rhiannonlynne:
I tried to do them in separate commits, but for some of them it was a bit too entangled.

PR #373, opened 2023-10-16
Tickets/opsim 1084

yoachim:
Update pre-computed DDF grid so it is sure to cover desired times.

### DIFF
--- a/rubin_sim/data/rs_download_data.py
+++ b/rubin_sim/data/rs_download_data.py
@@ -35,7 +35,7 @@ def data_dict():
         "movingObjects": "movingObjects_oct_2021.tgz",
         "orbits": "orbits_2022_3_1.tgz",
         "orbits_precompute": "orbits_precompute_2023_05_23.tgz",
-        "scheduler": "scheduler_2023_09_22.tgz",
+        "scheduler": "scheduler_2023_10_16.tgz",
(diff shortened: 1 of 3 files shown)


rhiannonlynne:
I think maybe the change to the "versions" got pushed directly to main that this PR is in support of?

rhiannonlynne:
I don't think we have this documented elsewhere ... 

the "ddf grid" generated by this code (and in the data downloads) contains what? The pre-scheduled DDF visits, starting at ?
Or a grid of potential times, but not the actual times? 
What is optimized when the scheduler reports "Optimizing DDF <XXX>" on startup? 
What happens if the survey start for a simulation is different than survey_start_mjd that is used to generate the DDFgrid?

rhiannonlynne (commented):
Would you mind putting some thoughts about what the DDF grid is or does in documentation somewhere? Could start out by reply to this PR and then transfer to the doc pages, which might be the lowest effort needed for the moment.

rhiannonlynne (approved):

rhiannonlynne:
For future reference .. 
The DDF grid file is a lookup table of the skybrightness/airmass/m5 depth over time for each of the DDF fields. 
The scheduler does the work of placing the DDF observations (as requested in the configuration) over time based on how many are requested, what m5 depth limits are in place, etc. -- although it does this for each DDF separately (so you can have collisions).

PR #374, opened 2023-10-17
update mjd0 to match v3.3 sims

yoachim:
Update the MJD0 value, also update the small sky brighntess file to match new mjd0 val.


rhiannonlynne:
When you rebase on main, the isort error should disappear, so ignore that. 
Looks good!

rhiannonlynne (approved):

PR #375, opened 2023-10-18
force wavelenth arrays to be idenical for twilight solar spectra

yoachim:
Quirks in `Bandpass` and `Sed`, so adding a line to prevent extra resampling of an `Sed` object.

### DIFF
--- a/rubin_sim/skybrightness/interp_components.py
+++ b/rubin_sim/skybrightness/interp_components.py
@@ -78,17 +78,21 @@ def load_spec_files(filenames, mags=False):
     The ESO npz files contain the following arrays:
     * filter_wave: The central wavelengths of the pre-computed magnitudes
     * wave: wavelengths for the spectra
-    * spec: array of spectra and magnitudes along with the relevant variable inputs.  For example,
-    airglow has dtype = [('airmass', '<f8'), ('solar_flux', '<f8'), ('spectra', '<f8', (17001,)),
-    ('mags', '<f8', (6,)]
+    * spec: array of spectra and magnitudes along with the relevant
+        variable inputs.  For example, airglow has dtype =
+        [('airmass', '<f8'), ('solar_flux', '<f8'), ('spectra',
+        '<f8', (17001,)),
+        ('mags', '<f8', (6,)]
 
-    For each unique airmass and solar_flux value, there is a 17001 elements spectra and 6 magnitudes.
+    For each unique airmass and solar_flux value, there is a 17001 elements
+    spectra and 6 magnitudes.
     """
 
     if len(filenames) == 1:
         temp = np.load(filenames[0])
         wave = temp["wave"].copy()
-        # Note old camelCase here. Might need to update if sav files regenerated
+        # Note old camelCase here. Might need to update if sav
+        # files regenerated
         filter_wave = temp["filterWave"].copy()
         if mags:
             # don't copy the spectra to save memory space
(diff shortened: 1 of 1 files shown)


rhiannonlynne (approved):

rhiannonlynne:
Where is the solar spectra coming from? (usually kurucz models aren't uniformly sampled, so I'm wondering where this is coming from).

rhiannonlynne:
I didn't dig through the code to see how this affects the problem Mike specified .. it seems like it's a cross-platform issue that was the cause of the problem for them (i.e. something like "1100 != 1100"??). I'm not entirely sure if this is going to avoid the problem ... and honestly, I'm not sure why it's happening. 
What's actually happening in the code is this: 
* the cannon and other filters are being read in and resampled onto a regular grid 
* the solar spectrum is being read in (and stays on its own grid)
* when the magnitude for the solar spectrum is calculated in any bandpass, it's resampled onto that regularly gridded bandpass (inside "calc_flux") if necessary (and here, machine precision may mean that it's called anyway). 
What's funny is that "calc_flux" calls Sed's resample_wavelength using the wavelength grid of the filter *already* .. so wavelength_grid should == wavelength_match and the solar spectrum should extend beyond the solar sed anyway. 

The way to really truly avoid the nan currently is to read in a solar sed that extends beyond the expected range of the wavelengths for the filters, including the Canon filter. What's odd, is this seems to be likely already true?
*** OH -- yeah, ok, so this is the problem (I finally unpicked the rest of the steps around this code). You're resampling the Canon filters onto the *solar* spectrum wavelength range, but you should just pick a useful range for these. .. why not 300 to 1100 nm? The solar sed ranges from 300.0 to 2499.999999999758 and it's likely the precision for this long-wavelength value that's causing the problem. 

I'm also going to put in a PR to change the "use a nan" to "use the specified fill value" and set that to default to 0. I'll also look around for some other places to simplify what we're doing, as it just got so complicated with extra calls when it was in use for catalogs. this will help prevent this in the future, but for now -- instead of resampling using a different wavelength step, please resample using a more limit wavelength *range*.

rubin_sim/skybrightness/interp_components.py
```diff
@@ -381,8 +381,11 @@ def __init__(self, mags=False, dark_sky_mags=None, fit_results=None):
             bp_temp.resample_bandpass(
                 wavelen_min=self.solar_spec.wavelen.min(),
                 wavelen_max=self.solar_spec.wavelen.max(),
-                wavelen_step=0.2,
+                wavelen_step=self.solar_spec.wavelen[1] - self.solar_spec.wavelen[0],
```
rhiannonlynne, line 384: The problem isn't here, it's that wavelen_min and wavelen_max should be within the range of the solar spectrum you're using. The step being 0.2 is fine.

rubin_sim/skybrightness/interp_components.py
```diff
@@ -381,8 +381,11 @@ def __init__(self, mags=False, dark_sky_mags=None, fit_results=None):
...
             )
+            # Force wavelengths to be identical so
+            # it doesn't try to resample again later
+            bp_temp.wavelen = self.solar_spec.wavelen
```
rhiannonlynne, line 388: If the solar spectrum had an irregular grid, this would be incorrect, so it's probably not a good example to set. (it's also worth noting that if I do "np.diff" on the solar spectrum wavelength array, I do get a few different values .. that look the same, but I'm assuming this machine precision stuff).

rhiannonlynne (changes requested):
It would be better and more robust to reduce the wavelength range for the Canon filters to stay within the wavelength range of the solar spectrum.

rhiannonlynne:
If you could please also run ruff and address the errors in this file before closing the PR, that would be great.

rhiannonlynne (approved):

PR #376, opened 2023-10-25
tickets/PREOPS-4124 Support restricting the scheduler reward summary stats to a region of interest

ehneilsen:
There are two approaches implemented here, which one might want to use in different circumstances.

First, one can use a mixin to turn a basis function that returns a healpy array into a scalar one.

Second, you can set an ROI on a survey, such that when the reward_df is made for that survey, only areas within the ROI are considered when finding the max, area, and feasibility.


yoachim (approved):
it has a unit test and everything!

PR #377, opened 2023-10-31
tickets/PREOPS-4501: Do not delete old SkyModelPre attributes until we know if new ones are valid

ehneilsen:
This probably doesn't come up in simulations, but can be a problem in interactive dashboards.


yoachim (approved):
Sure. Could also do `for attr in [self.sb, self.filternames, self.timestep_max]:` to save a few lines.

PR #380, opened 2023-11-06
Use the same code for basis function & reward area

ehneilsen:
The dataframe produced here used to do different things with scalar data when reporting available area. After this change, use common code for both basis function and accumulated reward area to avoid confusion, using the region of interest (where available).


yoachim (approved):
Looks good

PR #388, opened 2024-01-17
Add rubin-scheduler to all_req.txt

ebellm:
Currently the [Developer Installation](https://github.com/ebellm/rubin_sim/tree/u/ebellm/templates#developer-installation) instructions fail because `rs_download_data` now depends on `rubin-scheduler`, which is not included in `all_req.txt`.

### DIFF
--- a/all_req.txt
+++ b/all_req.txt
@@ -8,6 +8,7 @@ numexpr
 scipy
 sqlalchemy
 astropy
+palpy
 pytables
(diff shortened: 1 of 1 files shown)


rhiannonlynne (approved):
We should instead deprecate all_req.txt (and indeed, the new doc PR I have waiting there does that). But ok to merge this too.

PR #391, opened 2024-02-14
swap astropy for palpy

yoachim:
Remove last palpy dependency and use astropy for computing parallax factor.

### DIFF
--- a/all_req.txt
+++ b/all_req.txt
@@ -8,7 +8,6 @@ numexpr
 scipy
 sqlalchemy
 astropy
-palpy
 pytables
 h5py
 openorb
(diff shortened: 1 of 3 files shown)


rhiannonlynne:
Hi Peter, looks like GitHub is doing something unhappy about the caching again, which I will have a look at. Meanwhile, I'm assuming you ran this locally and it is fine? The results compare the same with palpy compared to astropy? 
Might be reasonable to drop in the PR a comparison of the calculation results with astropy versus palpy.

rhiannonlynne (approved):

yoachim:
image
Comparing this branch to `main`. Looks close enough to me!

PR #393, opened 2024-03-05
OPSIM 1132 - fix CI

rhiannonlynne:
Mostly update CI workflow. 
Added script to download subset of data, so it's easy to make small subset on GitHub. 
Moved get_baseline into rubin_sim (there will be corresponding PR on rubin-scheduler).


ehneilsen (approved):
Looks good to me.

PR #397, opened 2024-03-15
Simple metrics for density uniformity

ixkael:
This PR has three simple new metrics:
- a sky fraction metric (more flexible than existing ones)
- and edit to the totalpowermetric to include a remove_monopole 
- a new metric for linear combination of depth fluctuations

They come together in a sigma8 tomography metric which is demonstrated here: https://github.com/ixkael/ObsStrat/blob/meanz_uniformity_maf/code/meanz_uniformity/sigma8tomography_demo.ipynb

Annoyingly I didn't have time to figure out how to properly code this in a nice metric. This is partly because I couldn't think of an elegant way to avoid multiple calls to other existing metrics (including the new LinearMultibandModelMetric). For example the average depth is needed, and I wasn't sure how to avoid calling the ExgalM5 twice. The other problem is the need to use the depth maps N times (with N the number of redshift bins considered, here 5). Again I couldn't think of a nice way to run the metrics in a nested way in time for this deadline. But I hope the demo notebook illustrates things clearly enough, and I am happy to assist in any way I can.

### DIFF
--- a/rubin_sim/maf/metrics/summary_metrics.py
+++ b/rubin_sim/maf/metrics/summary_metrics.py
@@ -205,29 +205,59 @@ class ZeropointMetric(BaseMetric):
             return result
 
 
+class FskyMetric(BaseMetric):
+    """Calculate fraction of a desired footprint got covered.
+        To be applied as a summary metric to any healpix slicer.
+        Pixels with hp.UNSEEN or np.nan will be ignored, as well
+        as values falling outside of a ]min_val, max_val[ range.
+    Parameters
+    ----------
+    min_val, max_val : `float`
+        Minimum and maximum values for valid pixels (excluded)
+    """
+
+    def __init__(self, min_val=-np.inf, max_val=np.inf, **kwargs):
+        super().__init__(**kwargs)
+        self.min_val = min_val
+        self.mask_val = hp.UNSEEN
+        self.max_val = max_val
+
+    def run(self, data_slice, slice_point=None):
+        valid = data_slice["metricdata"] > self.min_val
+        valid &= data_slice["metricdata"] < self.max_val
+        valid &= data_slice["metricdata"] != self.mask_val
+        npix = data_slice["metricdata"].size
+        result = np.sum(valid) / npix
+        return result
+
+
 class TotalPowerMetric(BaseMetric):
     """
     Calculate the total power in the angular power spectrum between lmin/lmax.
     """
 
     def __init__(
-        self, col="metricdata", lmin=100.0, lmax=300.0, remove_dipole=True, mask_val=np.nan, **kwargs
+        self, col="metricdata", lmin=100.0, lmax=300.0, remove_monopole=True, remove_dipole=True, mask_val=np.nan, **kwargs
     ):
         self.lmin = lmin
         self.lmax = lmax
+        self.remove_monopole = remove_monopole
         self.remove_dipole = remove_dipole
         super(TotalPowerMetric, self).__init__(col=col, mask_val=mask_val, **kwargs)
 
     def run(self, data_slice, slice_point=None):
         # Calculate the power spectrum.
+        data = data_slice[self.colname]
+        if self.remove_monopole:
+            data = hp.remove_monopole(data, verbose=False, bad=self.mask_val)
         if self.remove_dipole:
-            cl = hp.anafast(hp.remove_dipole(data_slice[self.colname], verbose=False))
-        else:
-            cl = hp.anafast(data_slice[self.colname])
+            data = hp.remove_dipole(data, verbose=False, bad=self.mask_val)
+        cl = hp.anafast(data)
         ell = np.arange(np.size(cl))
         condition = np.where((ell <= self.lmax) & (ell >= self.lmin))[0]
         totalpower = np.sum(cl[condition] * (2 * ell[condition] + 1))
         return totalpower
+
 
 
 class StaticProbesFoMEmulatorMetricSimple(BaseMetric):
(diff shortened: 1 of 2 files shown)


rubin_sim/maf/metrics/summary_metrics.py
```diff
@@ -205,31 +205,61 @@ def run(self, data_slice, slice_point=None):
             return result
 
 
+class FskyMetric(BaseMetric):
```
rhiannonlynne, line 208: How is this different than the AreaThresholdMetric? And could we just modify the existing metric to add the extra capabilities, rather than introducing an entirely new version? (I would like to keep similar but slightly different code to a minimum, if possible).

rhiannonlynne: I'm also wondering about pixels with hp.UNSEEN or np.nan .. these *should* be getting masked in the metric_values masked array I would think?

rhiannonlynne (changes requested):
Some questions about the FskyMetric. 
remove_monopole looks good.

rubin_sim/maf/metrics/uniformity_metrics.py
```diff
@@ -0,0 +1,91 @@
...
+        units="mag",
+        extinction_cut=0.2,
+        min_depth_cut={'i': 25.0},
+        max_depth_cut={},
+        mean_depth={"u": 0.0, "g": 0.0, "r": 0.0, "i": 0.0, "z": 0.0, "y": 0.0},
+        n_filters=6,
+        **kwargs,
+    ):
+        """
+        Args:
```
rhiannonlynne, line 28: Could you please add docstrings in the LSST developer style (numpydoc style)? https://developer.lsst.io/python/numpydoc.html

rubin_sim/maf/metrics/uniformity_metrics.py
```diff
@@ -0,0 +1,91 @@
...
+        # add extinction_cut and depth cuts? and n_filters cuts
+        depths = np.vstack(
+            [
+                self.exgal_m5.run(data_slice[data_slice[self.filter_col] == lsst_filter], slice_point) - self.mean_depth[lsst_filter]
+                for lsst_filter in self.bands
+            ]
+        ).T
+        assert depths.shape[0] >= 1
+
+        if np.any(depths == self.badval):
```
rhiannonlynne, line 77: I'm not sure quite what you're testing for here .. self.badval isn't explicitly defined (but could be by a user), and it's not passed to exgal_m5 .. and the default is -666, which I don't think you're going to get from the exgal_m5 - self.mean_depth in general (and especially won't if the badval is defined differently for this metric than the exgal_m5 metric).

rubin_sim/maf/metrics/uniformity_metrics.py
```diff
@@ -0,0 +1,91 @@
...
+            return self.badval
+        for i, lsst_filter in enumerate(self.bands):
```
rhiannonlynne, line 79: It would be helpful to write this out with some more comment lines .. it's a bit opaque and not clear why you'd want to test for a maximum depth as well as minimum depth. (I'm assuming this is related to uniformity, but in a while from now .. we might not know).

rhiannonlynne:
I'll try to have a look at the notebook tomorrow, and see if I can say anything helpful!

rubin_sim/maf/metrics/summary_metrics.py
ixkael: thanks for pointing out AreaThresholdMetric, I hadn't seen it, but I now figured out how to use it, and thus I have removed the FskyMetric (see new commits). I have made two very small changes to AreaThresholdMetric though.

rubin_sim/maf/metrics/uniformity_metrics.py
ixkael: I addressed this in the new implementation

rubin_sim/maf/metrics/uniformity_metrics.py
ixkael: I've just added docstrings for most (if not all) functions/classes

rhiannonlynne:
Some nitpicks feedback (docstrings, line length, black, __all__'s, breaking things into new files) implemented in a PR on the fork - https://github.com/ixkael/rubin_sim/pull/1

rhiannonlynne:
It would still be good to add test coverage.

rubin_sim/maf/metrics/cosmology_summary_metrics.py
```diff
@@ -0,0 +1,985 @@
...
+
+        result["y1ratio"] = y1ratio
+        result["y10ratio"] = y10ratio
+
+
+class UniformAreaFoMFractionMetric(BaseMetric):
+    """
+    Run as summary metric on NestedRIZExptimeExgalM5Metric.
+
+    This metric StaticProbesFoMEmulatorMetric
```
rmandelb, line 765: Looks like this needs to be expanded?  Something like:

This metric uses maps of the combined RIZ exposure time and i-band depth maps (with a consistent set of area cuts) to identify potential reductions in cosmological constraining power due to substantial large-scale power in non-uniform coadds at a particular data release.  The RIZ exposure time map is used to identify whether there are residual rolling features.  If not, the metric returns 1.  If there are such features, then the region is segmented into similar-depth regions and the one with the largest cosmological constraining power is presumed to be used for science.  In that case, the metric returns the 3x2pt FoM (`StaticProbesFoMEmulatorMetric`, quantifying weak lensing and large-scale structure constraining power) for the largest of those regions, divided by that for the full region if it had been usable.

ixkael: Thank you for catching this gap Rachel! I committed and push this change.

PR #399, opened 2024-03-21
PREOPS-4949: Add stackers useful for schedview (Teff and dayobs related)

ehneilsen:
I'll do the addition of the plotters we discussed as a separate PR.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,6 +22,6 @@ repos:
         name: isort (python)
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.0.278
+    rev: v0.3.4
     hooks:
       - id: ruff
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ name = "rubin-sim"
 description = "Scheduler, survey strategy analysis, and other simulation tools for Rubin Observatory."
 readme = "README.md"
 license = { text = "GPL" }
-classifiers = [ 
+classifiers = [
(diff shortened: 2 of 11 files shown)


rhiannonlynne:
I realized I had a concern that if there were two bundles with TeffStacker but one "normed" and one not, that they might run at the same time and have a conflict about the t_eff column. However, apparently we built MAF better than that and it does the right thing (runs them separately, rewrites the t_eff values, and comes up with the correct answers. Yay! 

```
import rubin_sim.maf as maf

opsdb = '/Users/lynnej/opsim/baseline_v3.4_10yrs.db'

teff_stacker = maf.TeffStacker()
teff_stacker2 = maf.TeffStacker(normed=True)

bundle = maf.MetricBundle(maf.SumMetric(col='t_eff'), maf.UniSlicer(), '', stacker_list=[teff_stacker])
bundle2 = maf.MetricBundle(maf.MeanMetric(col='t_eff'), maf.UniSlicer(), '', stacker_list=[teff_stacker2])

g = maf.MetricBundleGroup({'teff': bundle, 'teff_normed': bundle2}, opsdb, verbose=True)
g.run_all()

bundle.metric_values, bundle2.metric_values
```
produces 
```
(masked_array(data=[44791395.476532385],
              mask=[False],
        fill_value=-666.0),
 masked_array(data=[0.6969573084883421],
              mask=[False],
        fill_value=-666.0))
```
as expected. 

values from the previous TeffMetric for the same run: 
All sky all bands	0.70	 (normed)   44791395.48 (total).

rhiannonlynne (approved):

PR #401, opened 2024-03-29
fix pandas dep warnings

yoachim:
Fixes several pandas deprecation warnings. Unit tests pass, and example notebook looks identical before and after update. Might not be compatible with older pandas versions.


rhiannonlynne:
I feel like there must be a more elegant way to do this than just copying the columns  but .. it works and it's clear that we're not making new data (which should make some other people happy) and it is likely fairly robust about pandas. I think this should still work with older versions. 
Are you getting (new) warnings about copying or making slices of the pandas data with this change?

rhiannonlynne (approved):

PR #402, opened 2024-04-01
Bugfix/opsim 1141

rhiannonlynne:
Fix Tissot ellipse plotting in BaseSkyMapPlotter.


yoachim (approved):
Looks good. The science radar seems to run with this.

PR #406, opened 2024-04-11
color and slope metrics

yoachim:
A pair of metrics that can be used as proxies for how well we will recover/classify transients like KNe.

### DIFF
--- a/rubin_sim/maf/metrics/color_slope_metrics.py
+++ b/rubin_sim/maf/metrics/color_slope_metrics.py
@@ -1,0 +1,188 @@
+__all__ = ["CheckColorSlope", "ColorSlopeMetric", "ColorSlope2NightMetric"]
+
+import numpy as np
+from rubin_scheduler.utils import int_binned_stat
+
+from .base_metric import BaseMetric
+
+
+class CheckColorSlope(object):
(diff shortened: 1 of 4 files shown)


rubin_sim/maf/metrics/color_slope_metrics.py
```diff
@@ -0,0 +1,188 @@
...
+        self,
+        mag=None,
+        night_col="night",
+        filter_col="filter",
+        m5_col="fiveSigmaDepth",
+        color_length=1.0,
+        slope_length=3.0,
+        time_col="observationStartMJD",
+        units="#",
+        metric_name="ColorSlope",
```
rhiannonlynne, line 96: I'm curious because I see this pattern a lot -- but why specify the metric_name here when it's the same as what will be constructed by default?

rhiannonlynne (approved):

rubin_sim/maf/metrics/color_slope_metrics.py
yoachim: I bet I was probably planning on having one metric with a kwarg rather then two, so I'd be setting the metric_name to run it different ways.

PR #409, opened 2024-04-29
OPSIM-1149: bugfix to add cols_req to new stackers

rhiannonlynne:
Relatively small changes, but important for making the stackers conform to the expected API. 

"cols_req" is a required attribute for stackers to run within the MAF framework. That is, in tests or if the data columns required are present anyway, these stackers will `.run` successfully, but in order for MAF to know what columns to pull from the database, the `cols_req` must be specified .. otherwise, the stacker will fail because the input columns won't necessarily be available. 

While fixing this, I also added a unit test that just calls "BaseStacker.help(doc=True)" as this will print all stacker cols_req and cols_added as part of generating its help string - this will ensure that all registered stackers have these attributes. 

I also updated some more ruff and docstring lines, and fixed a pandas future warning error (a simple fix, as the behavior we were previously doing matches the expected future behavior).


ehneilsen (approved):
Looks fine, and the expanded comments included in this PR will be quite useful as well.

PR #411, opened 2024-05-22
OPSIM-1155: Update DDF metrics / metric_subsets

rhiannonlynne:
In preparation for making a summary for the DDF working group, this PR 
* changes the name of "metric_set" to "metric_subset" 
* adds a Coaddm5PerNight metric
* supports UniSlicers being run with metrics which return a Histogram (object) value (because UniSlicers automatically have IdentityMetric added in order to add results to the summary database)
* Allows UniSlicers to carry slicePoint information (such as identifying the RA/Dec of a DDF location)
* Modifies an option for plotting with the OneDSlicer to permit accurate plotting of data with filled data values
* Quiets voluminous warnings about bin size from the "automatic bin size calculation utility"
* Updates DDF metrics to add CoaddM5PerNight. Also modifies some other cadence 'per night' metrics to run "Per DDF" instead of "at the center of the DDF" (i.e. using a UniSlicer + sql constraint, instead of UserPointSlicer at the center of the DDF).

### DIFF
--- a/rubin_sim/data/rs_download_data.py
+++ b/rubin_sim/data/rs_download_data.py
@@ -52,12 +52,12 @@ def data_dict():
     # to create tar files and follow any sym links, run: e.g.
     #  ``tar -chvzf maf_may_2021.tgz maf``
     file_dict = {
-        "maf": "maf_2022_08_26.tgz",
+        "maf": "maf_2024_06_13.tgz",
         "maps": "maps_2022_2_28.tgz",
         "movingObjects": "movingObjects_oct_2021.tgz",
         "orbits": "orbits_2022_3_1.tgz",
         "orbits_precompute": "orbits_precompute_2023_05_23.tgz",
-        "sim_baseline": "sim_baseline_2023_09_22.tgz",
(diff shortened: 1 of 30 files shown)


rhiannonlynne:
I guess this is likely to fail unit tests until we make a new tag for rubin_scheduler (to add the "get_current_footprint" function)

rhiannonlynne:
Ok - I needed to update the data download for MAF to add the metric_subsets.json file and a small h5 summary stat test file. 
At the same time, I added the data file for Eli's photometric calibration metric, so that'll be the next PR I think. 
Rubin-scheduler has been tagged so the get_current_footprint utility should be available. 
Let's see if there's anything else missing between dev and this environment.

yoachim (approved):
Looks all reasonable.

Seems like there's a `tests_2024_04_23.tgz` file that got created but isn't being used. We should update to using that one or delete it off sdf to avoid confusion.

rhiannonlynne:
Thanks - I thought the tests had been updated at some point, but I was just going with what was in the previous dictionary. Will update the tests to use the new version (which just drops the skybrightnessPre files from the tests directory) and double-check.

PR #412, opened 2024-05-31
Cleaned up assignment of self.coeffs to discard wasteful initilization.

bwSimpkins:
I am new to the LSST code community. Thank you all for the interesting work you all are doing here. I saw an extremely small change that would discard the neew to initialize self.coeffs in the cheby_values class. It doesn't changed the functionality of the set_coefficients but makes it more concise in my opinion.


yoachim (changes requested):
That makes it so `self.coeffs` is having keys set before it is defined. Note the unit tests fail.

rhiannonlynne:
Closing PR as the change causes tests to fail. 
Please feel free to open a new PR with other changes!

PR #415, opened 2024-07-18
switch for old ddf coords

yoachim:
add switch to use old ddf locations.


yoachim:
Sticking on a different PR

PR #417, opened 2024-08-06
U/yoachim/fo plot fix

yoachim:
Fix so FO plot is generated again. Add a switch so `ddf_batch` can be run using older coordinates.


rubin_sim/maf/batches/ddf_batch.py
```diff
@@ -486,7 +486,7 @@ def ddfBatch(
             for f in filterlist:
                 fieldsqls[f] = sqls[f]
         else:
-            fieldsql = f"note like '%{fieldname}%'"
+            fieldsql = f"scheduler_note like '%{fieldname}%'"
```
rhiannonlynne, line 489: This one we could use "target_name", but scheduler_note works too.

rubin_sim/maf/batches/glance_batch.py
```diff
@@ -144,7 +144,7 @@ def glanceBatch(
         bundle_list.append(bundle)
 
     # alt az of long gaps
-    sql = "note = 'long'"
+    sql = "scheduler_note = 'long'"
```
rhiannonlynne, line 147: Sigh, I see your point about scheduler_note vs note being embedded in many places. 
And true, a whole separate colmap just for this one change is likely overkill.

rhiannonlynne (approved):
Will tag rubin_scheduler so the necessary ddf_locations_pre3_5 is available.

PR #418, opened 2024-08-15
tickets/PREOPS-5360: clean up the nightsum and prenight Times Square notebooks by updating schedview to support them better

ehneilsen:


rhiannonlynne (approved):
I'm not super sure that I like the name "AgeMetric" just because it's hard to tell what that means, but I don't have a better suggestion. 
Seems fine.

PR #423, opened 2024-09-18
remove survey_start function

yoachim:
update from `rubin_scheduler` swapping `survey_start_mjd` to `SURVEY_START_MJD`.


yoachim:
Might need a new rubin_scheduler tag for the unit tests to pass.

ehneilsen (approved):
Looks good. I'm going to need to update schedview when this goes in.

PR #425, opened 2024-10-02
SP-1601: clean up rubin-sim developer guide issues

rhiannonlynne:
Humna and Lynne are working on improving rubin-sim adherence to developer code guidelines, also this lets us look at workflow.


tests/satellite_constellations/test_satellites.py
```diff
@@ -10,10 +10,10 @@ class TestSatellites(unittest.TestCase):
     def test_constellations(self):
         """Test stellite constellations"""
 
-        mjd0 = survey_start_mjd()
+        mjd0 = SURVEY_START_MJD
         sv1 = starlink_tles_v1()
-        sv2 = starlink_tles_v2()
-        ow = oneweb_tles()
+        _ = starlink_tles_v2()
+        _ = oneweb_tles()
```
yoachim, line 16: Rather than set the other tles to `_`, I would just loop the test over [sv1, sv2, ow]. Or if that's slow, check that those functions return something reasonable with something like `assert sv2 is not None`.

yoachim (approved):
Sad to see the comcam slicer go, but it's clearly bit-rotted at this point.

rhiannonlynne:
I was also sad to remove the comcam slicer given that it would be useful now! It was just not in a state that was usable. I think updating the camera footprint generation code and making a comcam camera map would be good.

tests/satellite_constellations/test_satellites.py
rhiannonlynne: Good suggestions, will update.

PR #427, opened 2024-10-21
Have sed.resample_sed actually resample rather than interpolate.

yoachim:
Replace numpy.interpolate calls with a resampling algorithm that conserves flux. Thus, `Bandpass` and `Sed` objects can now safely have very different wavelength sampling and still return correct integrated fluxes and magnitudes.

### DIFF
--- a/rubin_sim/phot_utils/__init__.py
+++ b/rubin_sim/phot_utils/__init__.py
@@ -3,3 +3,4 @@ from .photometric_parameters import *
 from .physical_parameters import *
 from .sed import *
 from .signaltonoise import *
+from .spectral_resampling import *
--- a/rubin_sim/phot_utils/bandpass.py
+++ b/rubin_sim/phot_utils/bandpass.py
@@ -74,17 +74,6 @@ class Bandpass:
             self.set_bandpass(wavelen, sb)
         return
 
-    def _check_wavelength_sampling(self):
-        """Check that the wavelength sampling is above some threshold."""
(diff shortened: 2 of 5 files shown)


rhiannonlynne:
Looks like this makes the magnitude calculation run about 2x slower .. not a dealbreaker I guess. 

The magnitudes that are calculated for the m5 'overview' values are the same, which is good .. either because the m5 is calculated using a flat SED and thus resampling methods don't matter as much or because the strict-regridding is happening behind the scenes anyway in the makeM5 function. Either way, having these results remain identical is good. 

The magnitudes calculated in other notebooks (such as the rubin_sim_notebooks.photometry directory) do change, although at the level of 0.001 magnitudes, but those are also using fairly simple SEDs. Did you get a chance to check the calculated magnitudes against some other package such as astropy synphot?

rhiannonlynne:
I did also check for the simple SED of the sun that if we read in the Bandpass or the SED and downsample either or both of them that we do still get about the same magnitude. Obviously you can still push things far enough that they are not accurate, but it's surprising how much leeway you do get, and you do regain some of the speed loss.

rhiannonlynne (approved):

yoachim:
Didn't check the mags against anything external, but looked like the values weren't changing.

I didn't see an easy way to get rid of the brute-force loop that's slowing things down. There is a numba wrapped version as well, so I'll make a note of that in the docstring in case anyone wants to upgrade to that in the future.

PR #428, opened 2024-10-28
tickets/SP-1685: handle unknown filters gracefully

ehneilsen:
consdb sometimes returns non-standard filters like "other". When encountered, don't crash.


rhiannonlynne (approved):
I am shocked this didn't come up in other contexts, but if this is the only place so far, it's fine.

ehneilsen:
> I am shocked this didn't come up in other contexts, but if this is the only place so far, it's fine.

This is the only place where it has failed for me so far. I won't be surprised if more instances crop up.

PR #430, opened 2024-10-31
update selfcal

yoachim:
Update `run_selfcal_metric` for `note` to `scheduler_note` schema update.
General updates to selfcal so filter is set properly.

### DIFF
--- a/rubin_sim/maf/maf_contrib/selfcal_uniformity_metric.py
+++ b/rubin_sim/maf/maf_contrib/selfcal_uniformity_metric.py
@@ -30,7 +30,14 @@ def _match(arr1, arr2):
 
 
 class PhotometricSelfCalUniformityMetric(BaseMetric):
-    def __init__(self, nside_residual=128, highglat_cut=30.0, outlier_nsig=4.0):
+    def __init__(
+        self,
+        nside_residual=128,
+        highglat_cut=30.0,
+        outlier_nsig=4.0,
+        metric_name="PhotometricSelfCalUniformityMetric",
+        filter_name="r",
+    ):
         cols = [
             "observationid",
             "fieldra",
@@ -39,7 +46,7 @@ class PhotometricSelfCalUniformityMetric(BaseMetric):
             "rotSkyPos",
             "filter",
         ]
-        super().__init__(col=cols, metric_name="PhotometricSelfCalUniformityMetric")
+        super().__init__(col=cols, metric_name=metric_name)
 
         filename = os.path.join(get_data_dir(), "maf", "monster_stars_uniformity_i15-18_sampled.parquet")
         self.stars = Table.read(filename)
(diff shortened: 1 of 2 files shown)


rubin_sim/maf/run_selfcal_metric.py
```diff
@@ -61,7 +72,7 @@ def run_selfcal_metric():
 
     # Make plots of the residuals map
     map_bundle = MetricBundle(
-        IdentityMetric(metric_name="PhotoCal Uniformity"),
+        IdentityMetric(metric_name="PhotoCal Uniformity, %s" % args.filter),
```
rhiannonlynne, line 75: Just to make things print nicer in the end, can you drop the ","? 
I admit this is just a matter of personal taste but  "Photocal Uniformity r <>" (and presumably maybe other bands ..)  just reads a bit easier than "Photocal Uniformity, r <>" ... 
I guess I don't care a ton but none of the other summary metrics use ","

rubin_sim/maf/run_selfcal_metric.py
```diff
@@ -43,13 +45,22 @@ def run_selfcal_metric():
...
     # Exclude DDF visits
-    sql = "note not like '%DD%'"
+    sql = "scheduler_note not like '%DD%'"
     # And run on only year 1 (?)
     sql += " and night < 366"
-    bundle = MetricBundle(selfcal_metric, slicer, sql, run_name=sim_name, info_label="year 1 no-DD")
+    sql += " and filter='%s'" % args.filter
+    # Exclude strange exposure times
+    sql += " and visitExposureTime > 20"
+    sql += " and visitExposureTime < 40"
```
rhiannonlynne, line 59: Do we actually need to drop these? It's true it makes things neater, but I would have thought including these exposures wouldn't make the self-calibration worse, and these visits ought to be calibrated as well .. 
I'd also wonder about including the DDF visits, tbh, but I think they were excluded just for calculation speed perhaps.

rubin_sim/maf/maf_contrib/selfcal_uniformity_metric.py
```diff
@@ -55,11 +62,10 @@ def __init__(self, nside_residual=128, highglat_cut=30.0, outlier_nsig=4.0):
         self.outlier_nsig = outlier_nsig
 
         self.units = "mmag"
+        self.filter_name = filter_name
 
     def run(self, data_slice, slice_point=None):
-        filter_name = data_slice["filter"][0]
```
rhiannonlynne, line 60: Well, that's interesting and I think I hadn't noticed. I don't think Eli had specified the constraint on the data either ... 
So, ha. Definitely should recheck the more downtime / less downtime outputs. 
I also wonder how well the metric tracks against the actual photometric calculation that Eli was expecting to do given that FGCM works across all filters (I think).

rhiannonlynne (approved):

rubin_sim/maf/run_selfcal_metric.py
yoachim: I was just being extra conservative, but they should be able to be used. That will let in the twilight observations, but I think you're right that they will be weighted properly.

PR #431, opened 2024-11-04
remove slow plots from batch

yoachim:
Remove plots from science batch that are very slow and not particularly informative.


rhiannonlynne:
I thought you'd changed these from plots where the points were drawn to the plots where the values were binned. This removes the plots entirely?

yoachim:
We could set them to `HealpixSkyMap`like the microlensing metrics are. I don't have a strong preference.

rhiannonlynne (commented):
Swap to healpixskymap

rhiannonlynne (approved):

PR #434, opened 2024-11-19
upadate sky generation

yoachim:
Small updates so generating pre-computed sky brightness files keeps working.


rhiannonlynne (approved):
Looks good. 
Why did the min have to be introduced where it wasn't needed before?

PR #439, opened 2025-01-29
SP-1906: backwards compatibility with numpy 1.

rhiannonlynne:
Copy input column names instead of optional copy.


ehneilsen (approved):
Looks fine to me.

PR #440, opened 2025-02-06
SP-1908: Update incremental template metric

yoachim:
Some minor updates to make the incremental template metric clearer.

### DIFF
--- a/rubin_sim/maf/metrics/incremental_template_metric.py
+++ b/rubin_sim/maf/metrics/incremental_template_metric.py
@@ -1,4 +1,5 @@
 import numpy as np
+from rubin_scheduler.utils import SURVEY_START_MJD, calc_season
 
 from .base_metric import BaseMetric
 
@@ -22,12 +23,14 @@ class TemplateTime(BaseMetric):
         Default 3.
     seeing_percentile : `float`, opt
         Maximum percentile seeing to allow in the qualified images (0 - 100).
-
-        Default 50.
+        XXX--deceptive kwarg since it checks the percentile at each point on the sky
+        and not continuously.
+        Default 100.
     m5_percentile : `float`, opt
         Maximum percentile m5 to allow in the qualified images (0 - 100).
-
-        Default 50.
+        XXX--deceptive kwarg since it checks the percentile at each point on the sky
+        and not continuously.
+        Default 100.
     seeing_col : `str`, opt
         Name of the seeing column to use.
     m5_col : `str`, opt
@@ -43,13 +46,14 @@ class TemplateTime(BaseMetric):
     def __init__(
         self,
         n_images_in_template=3,
-        seeing_percentile=50,
-        m5_percentile=50,
+        seeing_percentile=100,
+        m5_percentile=100,
         seeing_col="seeingFwhmEff",
         m5_col="fiveSigmaDepth",
         night_col="night",
         mjd_col="observationStartMJD",
         filter_col="filter",
+        mjd_start=None,
         **kwargs,
     ):
         self.n_images_in_template = n_images_in_template
@@ -65,6 +69,11 @@ class TemplateTime(BaseMetric):
             del kwargs["metric_name"]
         else:
             self.metric_name = "TemplateTime"
+        if mjd_start is None:
+            self.mjd_start = SURVEY_START_MJD
+        else:
+            self.mjd_start = mjd_start
+
         super().__init__(
             col=[self.seeing_col, self.m5_col, self.night_col, self.mjd_col, self.filter_col],
             metric_name=self.metric_name,
@@ -86,9 +95,9 @@ class TemplateTime(BaseMetric):
         bench_seeing = np.percentile(data_slice[self.seeing_col], self.seeing_percentile)
         bench_m5 = np.percentile(data_slice[self.m5_col], 100 - self.m5_percentile)
 
-        seeing_ok = np.where(data_slice[self.seeing_col] < bench_seeing, True, False)
+        seeing_ok = np.where(data_slice[self.seeing_col] <= bench_seeing, True, False)
 
-        m5_ok = np.where(data_slice[self.m5_col] > bench_m5, True, False)
+        m5_ok = np.where(data_slice[self.m5_col] >= bench_m5, True, False)
 
         ok_template_input = np.where(seeing_ok & m5_ok)[0]
         if (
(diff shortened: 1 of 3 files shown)


rubin_sim/maf/metrics/incremental_template_metric.py
```diff
@@ -102,7 +111,13 @@ def run(self, data_slice, slice_point=None):
         idx_template_inputs = ok_template_input[: self.n_images_in_template]  # Images included in template
 
         Night_template_created = data_slice[self.night_col][idx_template_created]
```
rhiannonlynne, line 113: While we're here -- this seems wrong, with a capital letter at the start of the variable?

rubin_sim/maf/metrics/incremental_template_metric.py
```diff
@@ -102,7 +111,13 @@ def run(self, data_slice, slice_point=None):
...
-        N_nights_without_template = Night_template_created - data_slice[self.night_col][0]
+        N_nights_without_template = Night_template_created
```
rhiannonlynne, line 114: This seems incorrect - n_nights_without_template should not be the same as the night_template_created.

rubin_sim/maf/metrics/incremental_template_metric.py
```diff
@@ -102,7 +111,13 @@ def run(self, data_slice, slice_point=None):
...
+
+        Frac_into_season_template_created = calc_season(
+            np.degrees(slice_point["ra"]),
+            data_slice[self.mjd_col][idx_template_created],
+            mjd_start=self.mjd_start,
+        )
```
rhiannonlynne, line 120: How is this the fraction of the season, if its' just calc_season? (calc_season will return the integer + fractional season)

rubin_sim/maf/metrics/incremental_template_metric.py
```diff
@@ -22,12 +23,14 @@ class TemplateTime(BaseMetric):
         Default 3.
     seeing_percentile : `float`, opt
         Maximum percentile seeing to allow in the qualified images (0 - 100).
-
-        Default 50.
+        XXX--deceptive kwarg since it checks the percentile at each point on the sky
```
rhiannonlynne, line 26: I really don't know if per-point-on-the-sky was the intended plan or not. I kind of think it was, so not necessarily deceptive but could be clarified (it also reads backwards to me) 
"Max percentile of seeing values (at each RA/Dec) to allow in the template source images (0-100)". 
It's also worth checking with Eric if the default should be 50 or 100. 
Previously we had a hard cutoff, which wouldn't be the worst way to go either (could potentially have both kwargs and use one to override the other).

rubin_sim/maf/metrics/incremental_template_metric.py
```diff
@@ -22,12 +23,14 @@ class TemplateTime(BaseMetric):
...
+        and not continuously.
+        Default 100.
     m5_percentile : `float`, opt
         Maximum percentile m5 to allow in the qualified images (0 - 100).
-
-        Default 50.
+        XXX--deceptive kwarg since it checks the percentile at each point on the sky
+        and not continuously.
+        Default 100.
```
rhiannonlynne, line 33: Likewise as above.

rubin_sim/maf/metrics/incremental_template_metric.py
rhiannonlynne: Pinging @ebellm for feedback on the update to the default value.

rubin_sim/maf/metrics/incremental_template_metric.py
rhiannonlynne: n_nights_without_template could be counted from the start of the survey, or it could be counted from the first observation in this field. Either would be valid. 
Or it could be dropped, given that "night_template_created" is present, as well as n_images_until_template".

rubin_sim/maf/metrics/incremental_template_metric.py
yoachim: Doing a percentile cut per-point-on-sky seems very wrong to me since then you could have 5 images with fantastic seeing, but it'll say it can't make a template because their aren't 3 over 50th percentile.

rhiannonlynne: I can see that .. and vice-versa if you have a spot on the sky with only terrible seeing, you might want to make a template anyway. 
I wonder if the actual way to do it is to mix the two together - if you have X images better than some (high percentile over the whole sky) value, make the template and if you have X images better than the 50% level at this point in the sky, make the template anyway as well.

ebellm: You are both right, this is tricky--you always want the best images, but what counts as "good enough" does have some dependence on the minimum airmass you can observe the field at, for instance.  I had though to use percentiles to avoid undue assumptions about the shape of the seeing distribution, but a hard cut to avoid the worst seeing tail + a local percentile might be a decent approach?

ebellm: Sounds fine to me.

rubin_sim/maf/metrics/incremental_template_metric.py
rhiannonlynne: Do you want to go ahead and drop it? I think this is reasonable.

rubin_sim/maf/metrics/incremental_template_metric.py
rhiannonlynne: I suppose it's only reported for the first year anyway, so it'll be fine.

rhiannonlynne:
To recap what I have done here: 
Instead of just calculating the percentile of the acquired visits over the (season? full survey?) and then using this alone to set the cutoff for m5/seeing at each slice_point (which led to questions about should the percentile be 100% which guarantees to take the first 3 visits, or if it should be 50% which might lead to waiting for quite a while .. and in operations we don't know this number anyway):
I now also take a default value for the zenith seeing at 500nm and a default dictionary for the m5 values (X=1.2, dark sky .. the standard m5 values from the throughputs).  At each slice_point, we can determine the best-possible airmass based on the slice_point declination and the site latitude, and then say what the corresponding expected threshold for the seeing (in the bandpass for the passed data, and accommodating the best-possible conditions) would be. Likewise, we can scale the m5 value to the declination of the slice_point (accommodating atmospheric extinction as a function of airmass and the seeing change from zenith, but not skybrightness changes).  Then the most permissive value of either the percentile seeing or the scaled seeing, and the percentile m5 or the scaled m5 is used. 

To recreate the behavior of either of the previous versions you can set the benchmark values to something ridiculously low (i.e. 0.2" for the seeing and m5 = {'u': 16, 'g': 16, 'r': 16, 'I': 16, 'z': 16, 'y': 16} and then set the percentile to whatever you wanted previously (50 or 100).

rhiannonlynne (approved):

PR #441, opened 2025-02-08
SP-1922: Bugfixes in scripts - set clobber properly, set order of batches

rhiannonlynne:
Thought I'd fixed these earlier, but this should now: 
* properly set no_clobber (to avoid repeating plots when we rerun metric batches)
* set the group name properly for show_maf webpages
* set the order of the metric batches consistently with what was previously online


yoachim (approved):
Looks fine

PR #443, opened 2025-02-13
updates to align more with DM guide

humnaawan:


rubin_sim/maf/web/maf_run_results.py
```diff
@@ -542,7 +542,7 @@ def order_plots(self, sky_plots):
             #  that do NOT match original _*_ pattern.
             pattern = "_[ugrizy]_"
             nonmatches = np.array([bool(re.search(pattern, x)) for x in sky_plots["plot_file"]])
-            nonmatch_sky_plots = sky_plots[nonmatches == False]
+            nonmatch_sky_plots = sky_plots[not nonmatches]
```
rhiannonlynne, line 545: Have you happened to run this script on this branch? 
I don't usually do that, but I am concerned that this may not work and perhaps should instead be 
`nonmatch_sky_plots = sky_plots[~nonmatches]`
.. but I'm not 100% sure which is why I think testing it might be the way to go. 
I don't think the unit test actually runs the script.

rhiannonlynne (approved):
Looks fine, but I'd like you to test if the `scimaf_dir` script actually works. 
In case you're not quite sure how to do that, once you are on this branch of your pip installed (with "-e") rubin_sim, you can navigate to any directory with an opsim database and just run `scimaf_dir --db <database filename>`.

rubin_sim/maf/web/maf_run_results.py
humnaawan: hi @rhiannonlynne i am not familiar with `scimaf_dir` but i ran it in a folder with a copy of the v4.1 baseline. the script ran for at least a few hours and im not sure how long the whole is supposed to take for me to check what you've suggested (and im assuming it actually has to finish and im still waiting on that).

also, i dont see a connection between [scimaf_dir](https://github.com/lsst/rubin_sim/blob/da85d111f0dbabf9ce70cffcbe374514c37b1757/rubin_sim/maf/scimaf_dir.py#L20) and [maf_run_results](https://github.com/lsst/rubin_sim/blob/da85d111f0dbabf9ce70cffcbe374514c37b1757/rubin_sim/maf/web/maf_run_results.py#L13) so im not sure how running `scimaf_dir` is testing the script.

humnaawan: okay indeed `scimaf_dir` didnt check the code - but it produced the file needed to run `run_maf` which did check the code. i got a 404 error with the change so ive pushed a fix now.

rhiannonlynne:
Yes, this looks good! Thank you.

PR #445, opened 2025-02-28
update SQL queries to proper quotes

yoachim:
Few lingering updates to SQL queries where single quotes are now being enforced.


rhiannonlynne (approved):

rhiannonlynne:
Ok, I approved this PR based on the contents of the PR, but something is leading rubin_scheduler to install v1.1.0 instead of 3.4.0 (which is why the unit tests are failing with 3.12). 
I'd like to understand why.

PR #451, opened 2025-03-20
tickets/SP-2004: move opsim simulation archive submodule out of rubin_scheduler into rubin_sim

ehneilsen:


batch/compile_prenight_metadata_cache.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+#SBATCH --account=rubin:developers      # Account name
+#SBATCH --job-name=auxtel_prenight_daily   # Job name
+#SBATCH --output=/sdf/data/rubin/shared/scheduler/prenight/sbatch/compile_prenight_metadata_cache.out # Output file (stdout)
+#SBATCH --error=/sdf/data/rubin/shared/scheduler/prenight/sbatch/compile_prenight_metadata_cache.err  # Error file (stderr)
+#SBATCH --partition=milano              # Partition (queue) names
+#SBATCH --nodes=1                       # Number of nodes
+#SBATCH --ntasks=1                      # Number of tasks run in parallel
+#SBATCH --cpus-per-task=1               # Number of CPUs per task
+#SBATCH --mem=16G                       # Requested memory
```
rhiannonlynne, line 10: I'm wondering about the configuration for this job. 
I think it's asking for 1 cpu, and the memory footprint for a single cpu on the milano nodes is generally pretty small (I think it's about 4GB). 
I wouldn't have expected this particular job to need 16GB, so if it doesn't, it seems like it would be easier for scheduling to have the requested memory be a bit smaller.

batch/run_prenight_sims.sh
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
+#SBATCH --account=rubin:developers      # Account name
+#SBATCH --job-name=auxtel_prenight_daily   # Job name
+#SBATCH --output=/sdf/data/rubin/shared/scheduler/prenight/sbatch/run_prenight_sims.out # Output file (stdout)
+#SBATCH --error=/sdf/data/rubin/shared/scheduler/prenight/sbatch/run_prenight_sims.err  # Error file (stderr)
+#SBATCH --partition=milano              # Partition (queue) names
+#SBATCH --nodes=1                       # Number of nodes
+#SBATCH --ntasks=1                      # Number of tasks run in parallel
+#SBATCH --cpus-per-task=1               # Number of CPUs per task
+#SBATCH --mem=16G                       # Requested memory
```
rhiannonlynne, line 10: Similar comment about the memory usage as for the compile metadata script.

rubin_sim/sim_archive/sim_archive.py
```diff
@@ -0,0 +1,1042 @@
...
+    max_simulation_age : `int`
+        The maximum age of simulations to consider, in days. Simulations older
+        than ``max_simulation_age`` will not be considered. Defaults to 2.
+    archive_uri : `str`
+        The URI of the archive from which to fetch the simulation.
+        Defaults to ``s3://rubin:rubin-scheduler-prenight/opsim/``.
+    compilation_uri : `str`
+        The URI of the compiled metadata HDF5 file for efficient querying.
+        Defaults to ``s3://rubin:rubin-scheduler-prenight/opsim/compiled_metadata_cache.h5``.
+
```
rhiannonlynne, line 965: I think it's helpful to have some brief explanation of what the kwargs can be or what they're used for. 
Even 
```
kwargs : `dict`
     kwargs are passed to rubin_sim.maf.get_sim_data, allowing selection of columns or constraints.
```
would give some guidance. 
(if kwargs are typically documented in a Notes section or something else, in numpydoc standard, I'd do it like that instead).

rubin_sim/sim_archive/sim_archive.py
```diff
@@ -0,0 +1,1042 @@
...
+    max_simulation_age: int = 2,
+    archive_uri: str = "s3://rubin:rubin-scheduler-prenight/opsim/",
+    compilation_uri: str = "s3://rubin:rubin-scheduler-prenight/opsim/compiled_metadata_cache.h5",
+    **kwargs,
+) -> pd.DataFrame:
+    """Fetches visit parameters from the latest archived pre-night simulation.
+
+    Parameters
+    ----------
+    day_obs : `str` or  `None`
```
rhiannonlynne, line 950: It's possible that for this particular call, day_obs being None is not useful. 
We could make day_obs required here (and make it clear whether this is the day_obs of the simulation creation or if it's the day_obs of the simulation .. ).

rubin_sim/sim_archive/sim_archive.py
```diff
@@ -0,0 +1,1042 @@
...
+
+def fetch_prenight_visits(
+    day_obs: str | None = None,
+    tags: tuple[str] = ("ideal", "nominal"),
+    max_simulation_age: int = 2,
+    archive_uri: str = "s3://rubin:rubin-scheduler-prenight/opsim/",
+    compilation_uri: str = "s3://rubin:rubin-scheduler-prenight/opsim/compiled_metadata_cache.h5",
+    **kwargs,
+) -> pd.DataFrame:
+    """Fetches visit parameters from the latest archived pre-night simulation.
```
rhiannonlynne, line 946: maybe not quite the right docstring description for this.

rubin_sim/sim_archive/sim_archive.py
```diff
@@ -0,0 +1,1042 @@
...
+        .join(f"{best_sim_date_index}", forceDirectory=True)
+        .join(sim_metadata[best_sim_uri]["files"]["observations"]["name"])
+    )
+    visits = get_sim_data(opsim_rp, **kwargs)
+
+    return visits
+
+
+def fetch_obslactap_visits(day_obs: str | None = None) -> pd.DataFrame:
+    """Return visits from best prenight briefing simulation.
```
rhiannonlynne, line 1011: I think the goal should be "Return visits from the best prenight briefing simulation for the next 24 hours" ? (upcoming night? next 24 hours is strictly speaking the requirement, but I think Wil is actually thinking of the upcoming night.. we should clarify how this should behave).

rhiannonlynne (changes requested):
Most of this looks good, but I think we have a tiny API detail to sort out between the `fetch_prenight_visits` and the `fetch_obsloctap_visits` that requires a clarification from Wil.

rhiannonlynne (approved):
Approved (following convo on slack about new obsloctap updates).

PR #452, opened 2025-03-27
tickets/SP-2082: Support getting obsloctac visits from the pre-night sim archive

ehneilsen:
I've reworked the start and end times of the obsloctap function to match what @rhiannonlynne proposed to Will:

> you call rubin_sim.sim_archive.fetch_obslactap_visits() at any time,
and we return the visits starting from the next sunset after the current time through the 2nd sunrise after that.
So, if you happened to call during the night (after sunset, before sunrise), we will NOT return any visits from the current night, but will return the visits from the next two nights.

### DIFF
--- a/rubin_sim/maf/utils/opsim_utils.py
+++ b/rubin_sim/maf/utils/opsim_utils.py
@@ -16,8 +16,8 @@ from sqlalchemy.engine import make_url
 
 def get_sim_data(
     db_con,
-    sqlconstraint,
-    dbcols,
+    sqlconstraint=None,
+    dbcols=None,
     stackers=None,
     table_name=None,
     full_sql_query=None,
@@ -79,10 +79,13 @@ def get_sim_data(
         table_name = "observations"
 
     if full_sql_query is None:
-        col_str = ""
-        for colname in dbcols:
-            col_str += colname + ", "
-        col_str = col_str[0:-2] + " "
+        if dbcols is None:
+            col_str = "*"
+        else:
+            col_str = ""
+            for colname in dbcols:
+                col_str += colname + ", "
+            col_str = col_str[0:-2] + " "
 
         query = "SELECT %s FROM %s" % (col_str, table_name)
         if len(sqlconstraint) > 0:
@@ -101,7 +104,7 @@ def get_sim_data(
             from lsst.resources import ResourcePath
 
             with ResourcePath(db_con).as_local() as local_db_path:
-                with closing(sqlite3.connect(local_db_path.ospath)) is con:
+                with closing(sqlite3.connect(local_db_path.ospath)) as con:
                     sim_data = pd.read_sql(query, con).to_records(index=False)
         except ModuleNotFoundError:
             raise RuntimeError(
--- a/rubin_sim/maf/utils/opsim_utils.py
+++ b/rubin_sim/maf/utils/opsim_utils.py
@@ -6,6 +6,7 @@ __all__ = (
 
 import os
 import sqlite3
+import urllib
 from contextlib import closing
 
 import numpy as np
@@ -79,13 +80,7 @@ def get_sim_data(
         table_name = "observations"
 
     if full_sql_query is None:
-        if dbcols is None:
-            col_str = "*"
-        else:
-            col_str = ""
-            for colname in dbcols:
-                col_str += colname + ", "
-            col_str = col_str[0:-2] + " "
+        col_str = "*" if dbcols is None else ", ".join(dbcols)
 
         query = "SELECT %s FROM %s" % (col_str, table_name)
         if len(sqlconstraint) > 0:
@@ -99,7 +94,7 @@ def get_sim_data(
     elif isinstance(db_con, str) and os.path.isfile(db_con):
         with closing(sqlite3.connect(db_con)) as con:
             sim_data = pd.read_sql(query, con).to_records(index=False)
-    else:
+    elif (not isinstance(db_con, str)) or urllib.parse.urlparse(db_con).scheme != "":
         try:
             from lsst.resources import ResourcePath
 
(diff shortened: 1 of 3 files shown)


rubin_sim/maf/utils/opsim_utils.py
```diff
@@ -77,26 +78,39 @@ def get_sim_data(
...
+        try:
+            from lsst.resources import ResourcePath
+
+            with ResourcePath(db_con).as_local() as local_db_path:
+                with closing(sqlite3.connect(local_db_path.ospath)) as con:
+                    sim_data = pd.read_sql(query, con).to_records(index=False)
+        except ModuleNotFoundError:
+            raise RuntimeError(
+                f"Cannot read visits from {db_con}."
+                "Maybe it does not exist, or maybe you need to install lsst.resources."
```
rhiannonlynne, line 112: I think this error message may need some tweaking. 
By far the most common error people are going to make is to try to use a mis-spelled version of the filename, or try to access a database that isn't where they think it is. 
Most of these people (who include not-us) will not be trying to load a file from an S3 bucket, but may not be savvy enough to realize that lsst.resources is irrelevant to what they need.  
The error message also slightly conflates two different issues - having lsst.resources versus the file existing anywhere (since it could still be a bad or inaccessible uri even with lsst.resources)

Can we tell if the uri is a local file vs. a uri that should be read with lsst.resources? And then give an error message that is as simple as "FileNotFoundError". 
If not,  then we could indicate why we were trying to use lsst.resources  .. 
```
try:
            from lsst.resources import ResourcePath
except ModuleNotFoundError:
            raise RuntimeError(f"Could not find {db_con} as a local file; please install lsst.resources in order to read remote file")
# then try to read the remote file
 with ResourcePath(db_con).as_local() as local_db_path:
                with closing(sqlite3.connect(local_db_path.ospath)) as con:
                    sim_data = pd.read_sql(query, con).to_records(index=False)
```

rubin_sim/sim_archive/sim_archive.py
```diff
@@ -931,3 +934,189 @@ def compile_sim_archive_metadata_cli(*args):
...
+    """
+
+    if first_day_obs is None:
+        first_day_obs = Time(Time.now().mjd - 0.5, format="mjd").iso[:10]
+    if last_day_obs is None:
+        last_day_obs = first_day_obs
+
+    sim_metadata = read_archived_sim_metadata(
+        archive_uri, num_nights=max_simulation_age, compilation_resource=compilation_uri
+    )
```
rhiannonlynne, line 985: is sim_metadata a pandas dataframe? would it make sense to make it one, so that the section below becomes a query of the dataframe?

rubin_sim/maf/utils/opsim_utils.py
```diff
@@ -77,26 +78,39 @@ def get_sim_data(
...
-        for colname in dbcols:
-            col_str += colname + ", "
-        col_str = col_str[0:-2] + " "
+        if dbcols is None:
+            col_str = "*"
+        else:
+            col_str = ""
+            for colname in dbcols:
+                col_str += colname + ", "
+            col_str = col_str[0:-2] + " "
```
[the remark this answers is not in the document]
ehneilsen: I was just moving what was there, without thinking about it too hard.
What's the list comprehension for? Why not just:

`", ".join(dbcols)`

rubin_sim/maf/utils/opsim_utils.py
ehneilsen: The error messages lists the two issues as possibilities because it can arise from either cause.

> Can we tell if the uri is a local file vs. a uri that should be read with lsst.resources?

I started to write code to do this, and concluded that it was more effort than it was worth, thinking that the user would know whether the file they are trying to read was local so diagnostics to be more specific wouldn't be particularly useful.

A test that is usually right shouldn't be that hard, though: if `db_con` is a string that does not start with a URI "scheme" designation, its *probably* intended to be a file name. But, in principle this can fail, because linux files names can have colons in them, so really any URI can be a path to a local file. Naming a file or directory `http:` or `s3:` is a foolish and confusing thing to do, but it can happen. It can also fail if support is added for new schemes.

rubin_sim/sim_archive/sim_archive.py
ehneilsen: sim_metadata is a list of dicts. It can be trivially turned into a dataframe, but doing that doesn't really simplify things here. Some of the elements we are concerned with are lists or dicts. Yes, a cell in a pandas DataFrame can be a dict or a list, but dealing with them is awkward, so using a DataFrame ends up making the code harder to read.

ehneilsen: I've reworked it so it only tries using ResourcePath if db_con is either not a str, or it can be parsed with pythons URL parser to get a valid scheme. So, it will only mention ResourcePath if it actually needs it.

rubin_sim/maf/utils/opsim_utils.py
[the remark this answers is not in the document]
rhiannonlynne: Oh true .. dbcols is already a list here. (was remembering this from a different situation, where the columns were coming back from a query to the consdb schema, and so was not a list).

rubin_sim/maf/utils/opsim_utils.py
rhiannonlynne: Yeah, ok - I'm ok if we can't decide whether it's local or not ourselves, but the thing to do is to make it clear that we're using lsst.resources for reading remote files only (rather than someone who didn't know what lsst.resources is thinking that they needed to install it to read their misspelled database filename). Because the general users for rubin_sim may very well not know what lsst.resources is or how to install it.

PR #453, opened 2025-03-27
show_maf refresh when tracking database file changes

yoachim:
Tested locally, this seems to make it so show_maf will pick up changes in the tracking database file after clicking anywhere.


rhiannonlynne:
Let's give it a try!

rhiannonlynne (approved):

PR #460, opened 2025-04-21
tickets/SP-2005: Add healpix and timestamp stackers in support of comparing completed to simulated visits

ehneilsen:


rubin_sim/maf/stackers/date_stackers.py
```diff
@@ -36,6 +38,79 @@ def _run(self, sim_data, cols_present=False):
...
+            case pd.Series():
+                sim_data["start_timestamp"] = empty_series
+            case np.ndarray():
+                sim_data["start_timestamp"] = empty_series.values
+            case _:
+                sim_data["start_timestamp"] = empty_series.values
+
+        return sim_data
+
+    def run(self, sim_data, override=False):
```
rhiannonlynne, line 82: I don't understand what you're doing here. 
The "run" method should call "_run", to actually add the data, but I don't see that here. 
It seems to me that this method to add the columns could override _add_columns in the base class instead?

tests/maf/test_stackers.py
```diff
@@ -287,6 +300,35 @@ def test_observation_start_datetime64_stacker(self):
...
+        value = pd.DataFrame(raw_value)
+        assert value.start_timestamp.dtype == "datetime64[ns, UTC]"
+        assert isinstance(value.start_timestamp[0], pd.Timestamp)
+
+        data = {"id": np.arange(num_points), "observationStartMJD": 61000 + 3000 * rng.random(num_points)}
+        stacker = stackers.ObservationStartTimestampStacker("observationStartMJD")
+        raw_value = stacker.run(data)
+        value = pd.DataFrame(raw_value)
+        assert value.start_timestamp.dtype == "datetime64[ns, UTC]"
+        assert isinstance(value.start_timestamp[0], pd.Timestamp)
```
rhiannonlynne, line 330: I think this test checks that the dtype of the values is correct, but does not actually test the values of the column. 
Please add a test that a computed value is correct.

rhiannonlynne (changes requested):

rubin_sim/maf/stackers/date_stackers.py
ehneilsen: My initial attempt at this was to override _add_columns, but that doesn't work, because the dtype I'm trying to add is a pandas thing, not a numpy thing, _add_columns only deals with numpy dtypes, and even if I added it as an "object" type in numpy, if I convert it back to pandas it still gets the dtype wrong. The simplest solution I could come up with is bypassing the _add_columns and _run mechanics entirely.

rhiannonlynne: Does this actually add the values though? I saw you had a "_run" method, but I don't see it being called from "run". 
I could just be confused -- if you're not actually calling "_run", does it need to be here?

ehneilsen: I think this was a mistake on my part -- I'll get rid of it. It's a leftover from when I was attempting to make this work with _run, before I gave up and just did everything in run.

rhiannonlynne (approved):
Thanks for the clarifications about the timestamp stacker- looks good.

PR #461, opened 2025-04-23
stellar mag fix

yoachim:
Had a numpy error and this fixed it. Kind of confused why the unit test didn't catch it sooner.


ehneilsen (approved):
Maybe the way you were doing it before was fine before the numpy update from 1 to 2?

PR #462, opened 2025-04-28
tickets/SP-2165: Set up automatic pre-night opsim simulations for lsstcam

ehneilsen:
This branch is itself a branch of the tickets/SP-2149 branch, which is itself a branch of the tickets/SP-2005 branch, so merging this will merge these others (if they are not independently merged first).

### DIFF
--- a/batch/run_prenight_sims.sh
+++ b/batch/run_prenight_sims.sh
@@ -29,6 +29,7 @@ conda activate /sdf/data/rubin/shared/scheduler/envs/prenight
 
 set -o xtrace
 
+export TELESCOPE=$1
 export AWS_PROFILE=prenight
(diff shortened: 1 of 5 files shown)


rhiannonlynne:
Ok - if this is a branch of a branch of other tickets (and I don't quite see how the stackers were relevant to the sims, nor how the logging was relevant to creating the option to build new sims for lsstcam) .. please clearly indicate what is new HERE and what features have been added in what files.

rhiannonlynne:
I think the changes here are the last three commits, and those seem like all they do is add a telescope value to the sim archive metadata and let that be set for whichever sim is being generated? 
This is likely fine. There could be other places where this is missed and I haven't noticed, and I suppose I should check unit test coverage at some point, but I also know we're likely to redo the sim archive metadata pieces in a bit with the postgres database, so it's probably fine for now.

rhiannonlynne (approved):

PR #465, opened 2025-05-15
tickets/SP-2211: Add rubin_sim.sim_archive.fetch_num_visits_in_latest_prenight_sim_for…

ehneilsen:
You can try it with:
```
import rubin_sim.sim_archive
rubin_sim.sim_archive.fetch_num_visits_in_latest_prenight_sim_for_night('2025-05-15')
```
or
```
import rubin_sim.sim_archive
rubin_sim.sim_archive.fetch_num_visits_in_latest_prenight_sim_for_night(20250515)
```

### DIFF
--- a/rubin_sim/sim_archive/sim_archive.py
+++ b/rubin_sim/sim_archive/sim_archive.py
@@ -12,9 +12,9 @@ __all__ = [
     "drive_sim",
     "compile_sim_archive_metadata_cli",
     "find_latest_prenight_sim_for_nights",
-    "fetch_latest_prenight_sim_for_nights",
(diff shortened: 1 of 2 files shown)


rhiannonlynne (changes requested):
See inline comment - a bit more ambitious about the API, but I think it'll last longer.

ehneilsen:
Okay, I've reworked the new function so we can enhance what we return later, and made the command name shorter.
Along the way to support this better, I refactored some of the functions it calls as well.

rhiannonlynne (approved):

PR #469, opened 2025-06-26
SP-2370: use band for DDF batch instead of filter

rhiannonlynne:
Only updated the ddf batch so far.
Also moved conda to optional requirements and updated required version of rubin_scheduler to match where sim_archive moved into rubin_sim.

### DIFF
--- a/rubin_sim/sim_archive/conda_packages.py
+++ b/rubin_sim/sim_archive/conda_packages.py
@@ -5,12 +5,12 @@ import sys
 from functools import lru_cache
 
 __all__ = [
-    "getCondaPackages",
+    "get_conda_packages",
 ]
 
 
 @lru_cache(maxsize=1)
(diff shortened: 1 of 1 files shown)


rubin_sim/sim_archive/conda_packages.py
```diff
@@ -0,0 +1,65 @@
...
+import sys
+from functools import lru_cache
+
+__all__ = [
+    "getCondaPackages",
+]
+
+
+@lru_cache(maxsize=1)
+def getCondaPackages() -> dict[str, str]:
```
ehneilsen, line 13: Can we follow PEP 8 and use snake case instead of camel case for functions?

ehneilsen (approved):
Except for the `getCondaPackages` function name not following PEP 8 (being camel case instead of snake case), this looks fine.

PR #472, opened 2025-07-15
SP-2453: Update rs_download_data to point to new movingObjects (2025-07-15) data

rhiannonlynne:
kurucz_sun was not scaled properly

### DIFF
--- a/tests/moving_objects/test_sed_data.py
+++ b/tests/moving_objects/test_sed_data.py
@@ -1,0 +1,61 @@
+import os
+import unittest
+
+from rubin_scheduler.data import get_data_dir
+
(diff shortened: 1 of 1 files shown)


rhiannonlynne:
Totally up for that, but do you mean just a unit test that unpacks the reflectance, multiplies by the kurucz_sun, and then calculates colors? Like in https://github.com/lsst/rubin_sim_notebooks/blob/main/photometry/solar_system_object_colors.ipynb but just for LSST?
(or I guess, at the start of the moving objects code, in order to calculate colors there)?

yoachim:
Sure, just reading in the files, doing something simple with them and checking that the values are reasonable.

PR #473, opened 2025-07-15
SP-2453: Use band OR filter in SRD batch and metrics

rhiannonlynne:
Replace data["filter"] with data[self.filter_col] in some places, and let sql query be more flexible in the SRD batch.


ehneilsen (approved):
Looks fine to me.

PR #474, opened 2025-07-18
filter to band update

yoachim:
Dangling "filter" to "band" fix. Maybe even backwards compatible.

### DIFF
--- a/rubin_sim/moving_objects/make_lsst_obs.py
+++ b/rubin_sim/moving_objects/make_lsst_obs.py
@@ -226,6 +226,11 @@ def make_lsst_obs():
         dbcols=None,
     )
 
+    # Check if we have "filter" or "band"
+    filter_col = "filter"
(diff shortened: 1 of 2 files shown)


rubin_sim/moving_objects/make_lsst_obs.py
```diff
@@ -226,6 +226,11 @@ def make_lsst_obs():
         dbcols=None,
     )
 
+    # Check if we have "filter" or "band"
+    filter_col = "filter"
+    if filter_col not in pointing_data.dtype.names:
+        filter_col = "band"
```
rhiannonlynne, line 232: This works, because of the colmap. 
For an old simulation, using the "fbs_sim" colmap ought to be backwards-compatible.

rhiannonlynne (approved):
This will not be the end of the filter-> band issue, I think, but glad to get further along! 
Looks good.

PR #516, opened 2026-04-06
tickets/SP-3021: make AnomalousOverheadFunc more flexible

ehneilsen:
The previous version just supported randomized changes to the scale, not a randomized offset. That is, the offset in the overhead was always proportional to the original overhead, while what we need currently is a random offset independent of the original overhead. This change adds support for the later, following any distribution supported directly by numpy's random number generator (although right now "normal" is all we need).

### DIFF
--- a/rubin_sim/sim_archive/prenight.py
+++ b/rubin_sim/sim_archive/prenight.py
@@ -15,8 +15,8 @@ import numpy as np
 import scipy.stats
 from matplotlib.pylab import Generator
 from rubin_scheduler.scheduler import sim_runner
-from rubin_scheduler.site_models import Almanac
 from rubin_scheduler.scheduler.utils.observation_array import ObservationArray
+from rubin_scheduler.site_models import Almanac
 
 from rubin_sim.sim_archive import make_sim_data_dir
 
@@ -212,9 +212,9 @@ class AnomalousOverheadFunc:
 
         # If we have the available parameters, make sure the final overhead
         # is more than min_overhead
-        overhead_checked : bool = False
+        overhead_checked: bool = False
         if obs is not None:
-            needed_data_found : bool = False
+            needed_data_found: bool = False
             try:
                 model_overhead = obs["slewtime"] + obs["visittime"] - obs["exptime"]
                 needed_data_found = True
@@ -225,7 +225,7 @@ class AnomalousOverheadFunc:
                 if after_offset_overhead < self.min_overhead:
                     overhead_offset = self.min_overhead - model_overhead
                 overhead_checked = True
-    
+
         if not overhead_checked:
             warnings.warn("Could not verify that the overhead was greater than the minimum.")
 
--- a/rubin_sim/sim_archive/prenight.py
+++ b/rubin_sim/sim_archive/prenight.py
@@ -9,9 +9,10 @@ import warnings
 from copy import deepcopy
 from datetime import date
 from pathlib import Path
-from typing import Callable
+from typing import Any, Callable, Mapping, Protocol
 
 import numpy as np
+import scipy.stats
 from matplotlib.pylab import Generator
 from rubin_scheduler.scheduler import sim_runner
 from rubin_scheduler.site_models import Almanac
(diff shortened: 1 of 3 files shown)


rubin_sim/sim_archive/prenight.py
```diff
@@ -78,13 +148,18 @@ def __call__(self, visittime: float, slewtime: float) -> float:
...
+        visit_overhead: float = visittime * self.rng.normal(self.visit_loc, self.visit_scale)
         # There might be anomalous overhead that makes visits take longer,
         # but faster is unlikely.
         if visit_overhead < 0:
             visit_overhead = 0.0
 
-        return slew_overhead + visit_overhead
+        scatter: float = self.scatter_dist_func(**self.scatter_kwargs)
+        overhead: float = max(
+            slew_overhead + visit_overhead + scatter, self.min_overhead - slewtime - visittime
```
rhiannonlynne: "overhead" is the non-on-sky time, right? 
So I think I get slew_overhead == slew time (or is it the extra scatter in the slew time? I am not clear, kind of looks like it might be 0), visit_overhead == extra visit time (which should always be 0, I can't even imagine what this means in our system if it's not 0), and scatter = how you've got a random scatter for each slew, which is somewhat how thing seems to work now for our real overhead, and then all together these are the overhead. 

But then I don't quite get why it's  min_overhead - slew time - visit time as the alternative .. is "visit time" part of the "overhead"? And if so, why isn't the visit time part of the visit_overhead? 

I wonder if including the visit time here as any part of this anomalous overhead function just makes things confusing. Is this supposed to return the actual "slew time" or is it the "slew time plus visit time"  (aka time between the start of successive images)?  What the return value is supposed to represent is confusing (and depends on the rubin_scheduler API, which is even more confusing).

rhiannonlynne (changes requested):
I think the documentation about what the return value should represent, for the callable needs to be improved.

rhiannonlynne:
Can you remind me why (if this should match the sim_runner / Scheduler API) this lives here and not in rubin_scheduler?

ehneilsen:
> Can you remind me why (if this should match the sim_runner / Scheduler API) this lives here and not in rubin_scheduler?

It is closely associated with running pre-night simulations, and so is part of the `sim_archive.prenight` submodule.
The `sim_archive.prenight` submodule has (or at least at some point had) dependencies that we were trying not to burden the observatory with. (Before the refactor of the archive stuff, the prenight routines dependend on the archive stuff because it handled not just running the sims but also adding them to the archive. This "harness" code has been moved to `lsst_survey_sim` now, so the motivation for keeping the `prenight` submodule with `sim_archive` is reduced.)

We _could_ split it out of the `AnomolousOverheadFunc` class and move it to `rubin_scheduler`, or move either the `sim_archive` or `sim_archive.prenight` submodules as a whole to `rubin_scheduler` (but this would add dependencies to `rubin_scheduler` we probably do not want there). Even moving either the `AnomolousOverheadFunc` class or the entire `prenight` submodule to `lsst_survey_sim` or moving either `sim_archive.prenight` or all of `sim_archive` own repository and python module are viable options.

My inclination, though, is to avoid adding or moving anything to `rubin_scheduler` that isn't absolutely required for observatory operations, and this is definitely not.

rubin_sim/sim_archive/prenight.py
ehneilsen: The "overhead" between two exposures = start_time_2 - start_time_1 - exptime_1

I revised the code (particularly variable names), comments, and docstring to make this clearer. Several variable names were indeed misleading: they hold the offsets from model overhead, not the total overhead itself, while the names of the variables implied the later.

At the same time, I also modified how visit overhead is being used, because a scatter in overhead that scales with the exposure time does not seem useful or interesting, and is not currently being used in any case.

rubin_sim/sim_archive/prenight.py
```diff
@@ -64,27 +127,67 @@ def __call__(self, visittime: float, slewtime: float) -> float:
             The visit time (seconds).
         slewtime : `float`
             The slew time (seconds).
+        exptime : `float`
```
rhiannonlynne: You are not using exp time below, and this modified the API for this function in rubin_scheduler.sim_runner ... can exp time be removed?

rhiannonlynne (approved):

PR #518, opened 2026-04-14
tickets/SP-3039: Move some functionality from schedview notebooks into rubin_sim where it can be reused

ehneilsen:
These additions to the sim archive code in rubin_sim move some of the code repeated across a couple of schedview notebooks to rubin_sim where it can be reused and tested.

While I was at it, I added tests for some of the sim archive functionality what was previously inadequately tested.
(The new testing was significantly LLM assisted. It tends to be very thorough.)

### DIFF
--- a/rubin_sim/sim_archive/prenightindex.py
+++ b/rubin_sim/sim_archive/prenightindex.py
@@ -154,7 +154,9 @@ def get_prenight_index(
     Parameters
     ----------
     day_obs : `str` or `int` or `date`
-        Night for which to fetch the index.
+        Night for which to fetch the index. The function
+        will return all simulations that simulate
+        the night designated by the day_obs.
     telescope : `str`
         Telescope name (default ``simonyi``).
     schema : `str` or `None`, optional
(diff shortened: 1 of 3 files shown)


rhiannonlynne:
The difference in intention between `get_sim_uuid` and `get_prenight_index` is unclear, at least partly because "index" is unclear. 
I *think* what get_prenight_index returns is a data frame with the UUID of the simulation + the resourcePath to the simulation output visits.  
This would be good to clarify in the docs. 

It would also be good to clarify the difference between the "dailyId" and the "index" and the "UUID". 

I am not sure I see how one would retrieve the UUID + resource path if you only had the day you wanted covered in the simulation and the kind of simulation you were looking for (out of the standard tags (assuming you did not want the nominal sim, but say, the sim with poor seeing). Are the day_obs, the creation date, and the tag information available from the get_prenight_index values?  .. so you could run `get_prenight_index` on the dayobs, filter it yourself to find the correct uuid? 

It would also be good to clarify what "day_obs" means in terms of prenight sims which typically cover 3 nights. Does this always correspond to the first night or any night within the simulation? 

I would like to see the order of arguments to `get_sim_index_info` reversed  (and could also refer to what is in the "index info" here too) -- the UUID is required, and is sufficient if you have access to the database; the day_obs is needed if you do not but only if you do not so could be optional. 
Reversing the order makes it clear that day_obs is not required if you have the database access available.

rhiannonlynne (approved):

PR #519, opened 2026-05-04
tickets/SP-3139: Add support for reading visits from hdf5 as well as sqlite3

ehneilsen:
Also added a flag to return `pandas.DataFrame`s to avoid needing to convert from `DataFrame`s to `recarray`s and back, which is slow.

### DIFF
--- a/tests/maf/test_opsimutils.py
+++ b/tests/maf/test_opsimutils.py
@@ -1,11 +1,13 @@
 import os
 import sqlite3
 import unittest
+from tempfile import TemporaryDirectory
 
 import numpy as np
 from rubin_scheduler.data import get_data_dir
 
 import rubin_sim.maf.utils.opsim_utils as opsimUtils
+from rubin_sim.sim_archive.util import opsimdb_to_hdf5
 
 TEST_DB = "example_v3.4_0yrs.db"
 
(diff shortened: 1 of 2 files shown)


tests/maf/test_opsimutils.py
```diff
@@ -75,6 +77,37 @@ def test_get_sim_data(self):
...
+            data_query = opsimUtils.get_sim_data(hdf5_file, full_sql_query=full_sql)
+            assert np.size(data_query) > 0
+            assert "observationId" in data_query.dtype.names
+            assert "fieldRA" in data_query.dtype.names
+            assert np.all(data_query["observationId"] < 5)
+
+            # Verify HDF5 results match SQLite results
+            data_sqlite = opsimUtils.get_sim_data(database_file, sqlconstraint="observationId < 10")
 
-if __name__ == "__main__":
```
yoachim, line 79: I like these lines since then one can run the unit test with `python test_filename.py`.

yoachim (approved):
Looks good to me

tests/maf/test_opsimutils.py
ehneilsen: These were accidentally removed, and I just added them back.

PR #523, opened 2026-06-09
tickets/SP-3223: Add extinction and cloud stacker

ehneilsen:
Two stackers are added to maf here, intended to be applied to visit sequences from consdb queries rather than opsim output.

The first uses sklearn robust linear fitting tools to fit for photometric extinction and zero point given the the zero_point_median, band, and airmass columns from consdb.
The next uses the results of this and zero_point_median to estimate the extinction from clouds, falling back on rubin_sim.phot_utils.predicted_zeropoint if there was no good fit. It supports an optional offset, so you can (and I probably will) feed in rubin_nights.reference_values.ZEROPOINT_OFFSETS_LSSTCAM, which if I understand what's going on will replicate what rubin_sim_addons does.

This was developed with a mixture of hand-coding and claude code assistance using the experimental support of claude code at the USDF.

### DIFF
--- a/rubin_sim/maf/stackers/extinction_stacker.py
+++ b/rubin_sim/maf/stackers/extinction_stacker.py
@@ -12,7 +12,6 @@ from rubin_sim.phot_utils import predicted_zeropoint
 from .base_stacker import BaseStacker
 
 logger = logging.getLogger(__name__)
-logger.setLevel(logging.DEBUG)
 
 
 def _make_default_band_priors():
@@ -72,7 +71,7 @@ def _limits_from_prior(prior):
     return k_min, k_max, zp_min, zp_max
 
 
-def _fit_extinction_one_group(
+def _fit_zp_extinction_one_group(
     airmass,
     zero_point,
     k_min,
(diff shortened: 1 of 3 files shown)


rubin_sim/maf/stackers/extinction_stacker.py
```diff
@@ -0,0 +1,500 @@
...
+        Keys are single-character band names; values are dicts with
+        keys ``k_min``, ``k_max``, ``zp_min``, ``zp_max``.
+    """
+    band_limits = {}
+    for band in "ugrizy":
+        # Extract k from the model by evaluating at two airmasses.
+        # The extinction coefficient is independent of exptime.
+        zp_at_X1 = predicted_zeropoint(band, airmass=1.0, exptime=1.0)
+        zp_at_X2 = predicted_zeropoint(band, airmass=2.0, exptime=1.0)
+        k_predicted = zp_at_X1 - zp_at_X2  # positive extinction coefficient
```
rhiannonlynne, line 36: Instead of this, we should just make the extinction coefficients available from the same phot_utils module as `predicted_zeropoint`.

rubin_sim/maf/stackers/extinction_stacker.py
```diff
@@ -0,0 +1,500 @@
...
+
+        # The fitted zeropoint is the intercept at airmass=0.
+        zp_at_X0 = zp_at_X1 + k_predicted
```
rhiannonlynne, line 37: Since the fit is to `k * (1 - X)` ... why do you want the zero point at X=0?

rhiannonlynne (commented):
ugh, realized I had some comments that have been stuck forever in pending. 
I also realize this is still a draft. 
Also, I suppose I don't remember the details here anymore (X-1) vs (X) vs (1-X)..

